PR #2, opened 2019-02-26
api v2

bddap:
This pr adds the ability to generate jsonrpc handlers for APIOwner and APIForeign via proceedural macro.

Since the handler generator is not specific to grin I've split it into its own crate, easy-jsonrpc. easy-jsonrpc provides a macro, jsonrpc_server, which decorates a trait definition and generates appropriate jsonrpc handlers.

This pr introduces a new trait, OwnerAPI, which the existing struct APIOwner implements. A new trait was added for several reasons:

1. APIOwner contains methods that don't make sense in a jsonrpc context; method signatures needed to be changed.
    - Mutable out parameters don't work in an rpc
    - Functions involving callbacks don't work in an rpc
    - Rpc methods can't return Result<, Error> because Error is not serializable. Result<, ErrorKind> is used instead
    - fn new() never needs to be called from a remote client
2. Adding a new trait is not a breaking change. Changing APIOwner to an rpc compliant trait would require refactoring other parts of grin-wallet.
3. APIOwner can trivially implement OwnerAPI (perhaps WalletBackend could implement it as well)

## TODO

- [x] Add doctests for every OwnerAPI method, OwnerAPI doctests will serve as jsonrpc documentation. See `OwnerAPI::accounts` for an example.
- [x] Add links to docs for APIOwner methods from each OwnerAPI method doc
- [ ] Implement fn initiate_tx(..) -> Result<(Slate, OutputLockFn<W, C, K>), Error> without using a callback
- [ ] Implement fn tx_lock_outputs(.. , mut lock_fn: OutputLockFn<W, C, K>) -> Result<(), Error> without using a callback
- [ ] Implement issue_send_tx like in refwallet/src/controller.rs
- [ ] Either A) Depreciate APIOwner in favor of OwnerAPI, or B) Rename OwnerAPI to be more decriptive e.g. RPCAPIOwner
- [ ] Add ForeignAPI trait
- [x] Add jsonrpc_client macro for typed client generation
- [ ] Possibly automatically generate clients in other languages such as javascript

## Blockers

This pr requires https://github.com/mimblewimble/grin/pull/2596, which implements serde for error types.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -400,10 +400,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "easy-jsonrpc"
-version = "0.2.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "easy-jsonrpc-proc-macro 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "easy-jsonrpc-proc-macro 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "jsonrpc-core 10.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.89 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -412,7 +412,7 @@ dependencies = [
 
 [[package]]
 name = "easy-jsonrpc-proc-macro"
-version = "0.2.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "heck 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -586,7 +586,7 @@ dependencies = [
 name = "grin_apiwallet"
 version = "1.1.0"
 dependencies = [
- "easy-jsonrpc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "easy-jsonrpc 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure_derive 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "grin_api 1.1.0 (git+https://github.com/bddap/grin?branch=milestone/1.1.0)",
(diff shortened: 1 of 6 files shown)


bddap:
This pr implements this [apiv2 proposal](https://github.com/mimblewimble/grin/issues/2425).

This implementation uses option 1. (see proposal) for error reporting. I tried to implement option 3., but it turns out [negative trait bounds](https://github.com/rust-lang/rust/issues/42721) would be required for differentiating result return types from non-result return types. Negative trait bounds are still a work in progress in rustc.

bddap:
Again, this pr relies on mimblewimble/grin#2596, so compilation will fail until that pr is accepted.

yeastplume:
Looks really good so far, very happy to see progress here and I'm going to keep checking out and testing locally as you go to provide more detailed feedback. Will have a think about the APIOwner naming.

I think Igno indicated that we'll be addressing https://github.com/mimblewimble/grin/pull/2596 once a couple of other things are in there.

yeastplume:
rust-secp has been updated now

yeastplume:
@bddap I've just completed some other chunks of work, so the V2 API is becoming the largest thing on my radar at the moment. Would you mind getting this to a state where it can be merged, then we can take that TODO list and turn it into a meta-issue and start tackling the work in parallel?

bddap:
@yeastplume sounds great. Glad to have help on this. I'll work on getting this to a mergeable state. Can you look into https://github.com/mimblewimble/grin/pull/2659? It needs to be merged before this.

graemes:
@yeastplume Would there be value in make the api compatible with the appropriate parts of the open banking api's (https://openbanking.atlassian.net/wiki/spaces/DZ/pages/23363964/Read+Write+Data+API+Specifications)  The banking sector in the UK is being forced to provide these apis and most other countries are heading down a similar route.  Could simplify adoption by wallet providers?

yeastplume:
@bddap https://github.com/mimblewimble/grin/pull/2659 was merged so let me know if there's anything else you need

@graemes We decided that using JSON-RPC is more appropriate than providing rest APIs, and I don't think the crypto space in general is too concerned with banking data standards (given there are many challenges unique to crypto that wouldn't be addressed by them) so I don't think we'd gain anything by deciding to adhere to them. If anyone felt strongly about it I'm sure they could wrap our JSON-RPC api to conform.

bddap:
@yeastplume this pr is ready to merge.

### Couple more things to do post-merge

Some OwnerApi/ForeignApi doctests are kinda BS at the moment. They expect errors because the NodeClient they are testing against is not running.
- [ ] Spin up a temporary NodeClient for OwnerApi/ForeignApi doctests.

Since we are updating APIs, now is a good time to step back and examine OwnerApi/ForeignApi methods. Which ones can we get rid of? Which ones should we change. @yeastplume you are probably well suited to that task.
- [ ] redesign network APIs if helpful

yeastplume:
@bddap Thanks, I'll merge this now then create another meta issue with all of the tasks

The doctests were written when everyone is in the same crate, and including them now involves circular dependencies which we can't have due to crates.io requirements. I'd actually started an empty node client at some stage to try to get around it, will leave the task up as a todo

Issue #6, opened 2019-03-07
slate serialization compatibility to 1.0.x


garyyu:
wallet send to 1.0.x wallet fail, and the error log on 1.0.x wallet:
```
Generic error: Invalid request body: invalid type: string "e1c141a8fce73ca7e8c546a865ba26ae1879259ab366fc54fd502bf0c04dc5fa", expected an array of length 32 at line 1 column 147 }
```

Looks like we didn't successfully compatible with 1.0.x slate serialization, but I remember we did.
To be checked...

garyyu:
Related PR: https://github.com/mimblewimble/grin/pull/2534

yeastplume:
I don't think there's actually code in there at the moment to produce and send the slate in v0 format. The slate versioning and conversion capability is there, but there's nothing that actually calls it.

garyyu:
Okay, so this is still a todo item for us before the formal release of this wallet :-)

mcdallas:
@garyyu This is working as intended, after 1.1.0 you won't be able to receive a tx from an old wallet if it was sent by a new wallet, only the other way around because the slate format changed to hex.

garyyu:
@mcdallas 
> This is working as intended

**Why we intend to have an incompatible with old version?**

But If this is a method to force upgrading, I fully agree 😄   because of https://www.grin-forum.org/t/critical-vulnerability-in-grin-1-0-1-and-older-fixed-in-1-0-2/4343

And the release note of `1.1.0` must highlight this incompatible and the purpose.

mcdallas:
Because how can the old client understand the new slate format? I don't think it's possible. I raised the issue of backwards incompatibility in the dev meeting and people seemed fine with it.

yeastplume:
Just thinking about the work involved in maintaining compatibility between all wallet versions.

Say there are 4 versions of the slate, v1, v2, v3 and v4.

If you're on V4, you can "upgrade" all incoming messages from v1, v2 and v3 clients. However, to respond to them you also need to be able to downgrade to their message format, so I need to maintain code to upgrade from v1 -> v2 -> v3 -> v4 as well as to downgrade v4 -> v3 -> v2 -> v1. This will have to be an incremental process, so in adding v5 I need to add upgrade v4 -> v5 and downgrade v5 ->v4 and continue down the chain depending on who I'm talking to. So for each element in each version of the slate there are From<> implementations that upgrade to the next and downgrade to the previous.

[...]

PR #13, opened 2019-03-11
Slate version conversions

yeastplume:
Bit messy thus far, but trying to come up with a scheme that will let the wallet upgrade and downgrade slate versions as needed in as transparent a manner as possible. Each slate version can upgrade or downgrade itself to the next/previous versions, while the latest version should be able to deserialise into the current Slate.

https://github.com/mimblewimble/grin-wallet/issues/6


yeastplume:
Updated now. Newer wallets should be able to receive slates from older wallets, convert them to the latest format, then downgrade them back to the earlier format to send them back. A few version fields were added to the V2 slate struct to facilitate this.

Also added a new command line arg for the send command that allows user to target an earlier slate version, if they absolutely have to send a slate to an older wallet.

PR #17, opened 2019-03-13
default to using a Plain kernel with no lock_height

antiochp:
We currently default to generating `HeightLocked` transaction kernels (locked at current block height).
This PR changes this to default to generating `Plain` transaction kernels.

We are exploring making the lock_height optional on kernels (it is not included in the tx kernel hash for `Plain` kernels).
First step toward making them truly optional is to stop using them where we don't need to. 

Eventually we will be able to reduce storage and bandwidth costs (a little) by only including the `lock_height` on `HeightLocked` kernels.

This PR leverages the existing grin behavior - a `Plain` tx kernel default to a `lock_height` of 0 (as we need something to store in the u64 lock_height). This is excluded from the kernel hash for `Plain` kernels.

TODO - 
- [ ] Some reasonably extensive testing to make sure `Plain` kernels are usable across all wallet and node scenarios.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,5 @@ target
 grin.log
 wallet.seed
 test_output
-wallet_data
-wallet/db
+wallet*
 .idea/
--- a/controller/tests/transaction.rs
+++ b/controller/tests/transaction.rs
@@ -18,6 +18,7 @@ extern crate grin_wallet_controller as wallet;
 extern crate grin_wallet_impls as impls;
 extern crate grin_wallet_libwallet as libwallet;
 
+use self::core::core::transaction;
 use self::core::global;
 use self::core::global::ChainTypes;
 use self::keychain::ExtKeychain;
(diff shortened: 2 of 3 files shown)


antiochp:
@tromp Will be pleased to know about this I suspect.

tromp:
We do include lock_height in the kernel of HeightLocked transactions, as required to prevent malleability to different lock_heights...

And yes, I'm very happy to see txs defaulting to plain!

antiochp:
> We do include lock_height in the kernel of HeightLocked transactions, as required to prevent malleability to different lock_heights...

Yes we do. I realized my description above was somewhat ambiguous, revised to make it clearer.

yeastplume:
:+1:, given 1.1.0 is going to require quite a bit of testing anyhow, do you want to merge this now? Or do you want to test separately then merge?

antiochp:
Let me fix up the conflicts and I'll merge this in.
I can run this locally against 1.0.x nodes and it all seems to be fine.
We should keep this change in mind though when testing 1.1.0 (in case anything unexpected crops up related to this).

PR #21, opened 2019-03-14
feat(ci): introduce code coverage

mrekucci:
The build result is determined as soon as all the required jobs
finish, while the "coverage" job may continue and may even fail.


yeastplume:
No longer relevant since switch to azure

PR #23, opened 2019-03-15
APIV2 refactoring

yeastplume:
Introducing the v2 api is demonstrating some places where the separation of concerns between crates could be much better, so taking the opportunity to do a refactoring. Will update this with a description of the new crate structure and the reasoning behind it before merging.


yeastplume:
Okay, so what's happened here is:

* Split up the previous `refwallet` into 2 crates, currently called `impls` and `controller`. `impls` mostly contains concrete implementations of the traits found in `libwallet`. `controller` just contains logic to handle command line args and control the listener APIs. It was necessary to split these up to avoid circular dependencies in the `api` crate, which will need to use the types in `impls` to create self-testing rpc api doc tests.
* Move the implementations of the owner and foreign APIs into `libwallet`. All the api layers do now is unlock/open the wallet before calling the implementations. This was needed to allow internal code to call api functions directly, again without creating circular dependencies.

PR #35, opened 2019-04-02
Slate compatibility between versions

yeastplume:
Changes as a result of trying to send slates between different wallet versions. In the 1.1.0 code, all slates are serialized and sent as strings, to allow for differences between versions. Previously, they were serialized directly into a slate object, and the representation comes out differently after serde deser.

Bit of messiness has to go in to compensate for this for the time being (until 1.0.x series becomes completely obsolete), but for now the cases are handled and the following interactions should work:

* Sending a transaction from 1.1.0 to 1.1.0
* Sending a transaction from 1.0.x to 1.1.0 
* Sending a transaction from 1.1.0 to 1.0.x, provided the -v [target_slate_version] flag is specified in the send command (set this to 0 for compatibility with all older wallet clients)

### DIFF
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -904,16 +904,14 @@ where
 	T: Serialize,
 {
 	match serde_json::to_string(s) {
-		Ok(json) => {
-			response(StatusCode::OK, json)
-		}
+		Ok(json) => response(StatusCode::OK, json),
 		Err(_) => response(StatusCode::INTERNAL_SERVER_ERROR, ""),
 	}
 }
 
 // As above, dealing with stringified slate output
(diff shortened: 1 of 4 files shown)


lehnberg:
Does this mean that `grin wallet send ... -v 0` ensures compatibility across any client (inc. v1.1.0)? Or is this for compatibility with <1.1.0 clients only?

yeastplume:
Yes, `-v 0` should work with any existing client, including v1.1.0. We're going to need to start thinking about upgrade policies though, one reason being that I don't think we want to create a state where everyone just uses -v 0 by default for the rest of time.

PR #89, opened 2019-04-23
fix #88

### LINKED ISSUE #88: API V2 handlers treat jsonrpc notifications as errors.
JSON RPC spec includes a [notification](https://www.jsonrpc.org/specification#notification) feature. Requests marked as notifications are handled silently; the method is run, but no response is sent back.

easy-jsonrpc, the crate we use for APIV2, supports notifications. When [handle_request](https://docs.rs/easy-jsonrpc/0.4.1/easy_jsonrpc/trait.Handler.html#method.handle_request) receives a notification, it handles the requests normally, but returns `None` to indicate no reply is necessary.

### Problem:

Grin-wallet's controller logic handles notifications as errors [here](https://github.com/mimblewimble/grin-wallet/blob/9b49bde4f91e642779327d03ac7775cfe30b5c33/controller/src/controller.rs#L689-L695) and [here](https://github.com/mimblewimble/grin-wallet/blob/9b49bde4f91e642779327d03ac7775cfe30b5c33/controller/src/controller.rs#L868-L874).

### Solution:

When `handle_request` returns `None`, reply with `json!(null)`.

bddap:
Fixes issue #88 by replying to json rpc notifications with an empty list.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -402,10 +402,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "easy-jsonrpc"
-version = "0.4.1"
+version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "easy-jsonrpc-proc-macro 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


bddap:
Also bumps easy-jsonrpc to 0.5.1 for this [usability fix](https://github.com/layer1capital/easy-jsonrpc/issues/5).

yeastplume:
Great, thanks for this!

PR #96, opened 2019-05-02
Command line implementation of invoice commands

yeastplume:
Following from #90, the actual command line implementation exercising the invoice commands.

The code will mostly be a lot of boilerplate copying from the parsing and api execution system that's in place, but the particular file to watch for comments on the UI (if you can even call it that) is the command line arg yaml definition file:

`src/bin/grin-wallet.yml`

- [x] invoice command
- [x] pay command
- [x] finalize  command
- [ ] community review, particularly on naming, arguments, and process_invoice workflow
- [x] test implementation exercising new commands (ongoing)

### DIFF
--- a/controller/src/command.rs
+++ b/controller/src/command.rs
@@ -421,6 +421,104 @@ pub fn issue_invoice_tx(
 	Ok(())
 }
 
+/// Arguments for the process_invoice command
+pub struct ProcessInvoiceArgs {
+	pub message: Option<String>,
+	pub minimum_confirmations: u64,
+	pub selection_strategy: String,
+	pub method: String,
+	pub dest: String,
+	pub max_outputs: usize,
+	pub target_slate_version: Option<u16>,
+	pub input: String,
+	pub estimate_selection_strategies: bool,
+}
+
+/// Process invoice
+pub fn process_invoice(
+	wallet: Arc<Mutex<WalletInst<impl NodeClient + 'static, keychain::ExtKeychain>>>,
+	args: ProcessInvoiceArgs,
+	dark_scheme: bool,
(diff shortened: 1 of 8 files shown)


yeastplume:
Added `issue_invoice` command as a start. The `process_invoice` command is next, and it's likely to be the most difficult due to the need for interaction.

mcdallas:
I would propose changing `issue_invoice` to `invoice` (as in the verb), `process_invoice` to `pay` and maybe use the existing `finalize`.

Also it would be nice if by default the invoice would be printed to stdout as a base64 string (instead of file) like in the Lightning network. The workflow I am thinking is:
- You checkout from a site, they give you a base64 string plus a textbox to paste your reply.
- User copies the string and types in his wallet `grin-wallet pay <string>`
- "Do you want to pay X for Y amount (yes/no)"
- Wallet prints another string
- User pastes it to the textbox and clicks submit

DavidBurkett:
I like @mcdallas's suggestion. Seems like a simple enough workflow for most users, and the wording makes sense to me.

lehnberg:
👍

yeastplume:
Fine will all of this, except using the same `finalize` as the regular send case, as the rest of the code keeps them separate throughout. Will make changes

yeastplume:
Think this is just about there, couple of points:

* I went with reusing the same finalize command and have the command determine whether the Slate is an invoice or not.
* Will leave the base 64 stuff as a future enhancement. If we start outputting base64 we'd want it consistent across all wallet commands, so it most likely deserves its own thinking and PR.

yeastplume:
Merging to allow builds for PR work, sure we can still review and comment on the beta build

PR #103, opened 2019-05-13
Simplify slate (de)serialization

jaspervdm:
- Previous deserialize method did multiple full deserializes. I added a `SlateVersionProbe` which only has to deserialize the relevant fields to figure out the version, after which full deserialization can be done.
- Implement `Serialize` on `Slate`, which serializes to the orig version. Use it in the adapters and the foreign API
- Simplify HTTP adapter, no longer require multiple serializes and deserializes

Still some testing to do, so marking as draft. Requesting review from @yeastplume

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -280,16 +280,16 @@ dependencies = [
 
 [[package]]
 name = "croaring"
-version = "0.3.8"
+version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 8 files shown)


yeastplume:
Looking very good, and thanks for cleaning up the cruft here.

Are you ready to merge it in? I'd like to get one more beta release done before the actual 1.1.0 release.

jaspervdm:
@yeastplume I will do some final testing and let you know in a few hours

jaspervdm:
Did some testing with sending/receiving slates, fixed a bug in the HTTP adapter. Ready for final review

Issue #108, opened 2019-05-17
An error occurred "Wallet command failed: LibWallet Error: Can't Deserialize slate"


llgoer:
Send by http `./grin-wallet send -m http -d http://x.x.x.x:3415 76`
An error occurred
```
Password: 
Wallet command failed: LibWallet Error: Can't Deserialize slate
```
I just in 'chain_type = "UserTesting"'

send by file success.

I use latest beta version of grin wallet.

grin:v1.1.0-beta.2
grin wallet : v1.1.0-beta.3

vandabbin:
I am using grin:v1.1.0 and grin-wallet:v1.1.0 and getting the same 
`Wallet command failed: LibWallet Error: Can't Deserialize slate`
when attempting to send to an https address. works fine when saved to file

Incrediblez7:
Please reopen it. I'm having the same issue. 
`grin-wallet send -m http -d https://grin.hotbit.io/353758 1.64`
The exact same error appeared: `Wallet command failed: LibWallet Error: Can't Deserialize slate`

Grin: 1.1.0
Grin-wallet: 1.1.0

kargakis:
Issue reopened at https://github.com/mimblewimble/grin-wallet/issues/145

pro2pop:
hi! the same  trouble 
 grin-wallet send -m http -d "https://grin.hotbit.io/315748" 10
Password:
Wallet command failed: LibWallet Error: Can't Deserialize slate
node 1.1.1-beta.1
wallet 1.1.0

taoteh1221:
Same issue in v1.1.0. Additionally, when trying the optional method of:
``
grin-wallet send -d "transaction.tx" -m file XX.XX
``
although the CLI response seems ok:
``
Command 'send' completed successfully
``
Poloniex rejects the file:
``
Invalid File
``

jaspervdm:
The are most likely running pre v1.1 wallet, so use `-v 0` flag when sending.

taoteh1221:
Thanks for the idea @jaspervdm , however even this was rejected by poloneix as "invalid file":
``
grin-wallet send -d "transaction1.tx" -v 0 -m file 1.00
``
Maybe I should try -v 1?

taoteh1221:
Looks like you HD seed in v1.1.0 isn't backwards compatible with v1.0.0 as well (seems that way for me at least, idk). I'm presuming funds are stuck until v1.1.0 is fixed.

[...]

PR #114, opened 2019-05-20
Remove double json encoding on http response

jaspervdm:
I tested all permutations of sending/listening with v1.0.3 and v1.1.0 using HTTP and file. Turns out one of them wasn't working due to a double serialization, which is removed with this PR.

Fixes https://github.com/mimblewimble/grin-wallet/issues/108


yeastplume:
Great, and thanks for testing all of these permutations

PR #121, opened 2019-05-27
Fix wallet for non-Copy SecretKey

eupn:
This PR makes grin-wallet able to compile in case `secp::SecretKey` made non-`Copy` by this PR: https://github.com/mimblewimble/rust-secp256k1-zkp/pull/51

Do not merge it unless abovementioned PR is merged so things won't break.

*Update*: https://github.com/mimblewimble/rust-secp256k1-zkp/pull/51 is merged, so this PR can be merged as well.

### DIFF
--- a/libwallet/src/slate.rs
+++ b/libwallet/src/slate.rs
@@ -483,7 +483,7 @@ impl Slate {
 		let blind_offset = keychain.blind_sum(
 			&BlindSum::new()
 				.add_blinding_factor(BlindingFactor::from_secret_key(sec_key.clone()))
-				.sub_blinding_factor(self.tx.offset),
+				.sub_blinding_factor(self.tx.offset.clone()),
 		)?;
 		*sec_key = blind_offset.secret_key(&keychain.secp())?;
 		Ok(())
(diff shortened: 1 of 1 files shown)


garyyu:
thanks @eupn to handle this 👍  

I think this 3 small changes (from copy to clone) doesn't depend on low level rust-secp256k1-zkp.

And are these 3 places the all happenings of SecretKey copy?

For example:
```
		Context {
			parent_key_id: parent_key_id.clone(),
			sec_key: sec_key,    <<<<  is this a 'copy' or a 'clone'?
			...
```
And:
```
		*sec_key = blind_offset.secret_key(&keychain.secp())?;
```
Just 2 examples here, I believe there're more like these.
(and same question for grin server, I think the better way to check this is to use the latest  `rust-secp256k1-zkp` master to check)

How?
In grin repo util/Cargo.toml
```
[dependencies.grin_secp256k1zkp]
path = "../../rust-secp256k1-zkp"     <<<< uncomment this line and put your rust-secp256k1-zkp there.
version = "0.7.5"
```

yeastplume:
Thanks, targeting for 2.0.0 and will review more closely after 1.1.0 is released

garyyu:
Related:  https://github.com/mimblewimble/grin/pull/2847#issuecomment-499268857

yeastplume:
Apologies, I went and fixed this while creating 1.1.0, forgetting this PR was here. Should be good now, thanks!

PR #126, opened 2019-05-30
Changes for 1.1.0 -> 2.0.0 Deployment strategy decisions

yeastplume:
1.1.0 Changes resulting from dev meetings over the past couple of days. All points are outlined below, but checklist for this PR only applies to 1.1.0.

Also, in addition to checklist below:

* Clean some compiler warnings during test
* Change misleading output saying that slate was sent when it was actually just being output to file
* Update API doctests to reflect signature format change and output features change grin-side

- [x] Grin-wallet v1.1.0 will default slate output to SlateV0, and ensure that users who want to output the newest SlateV2 format must add -v 2 to their command. (Note that invoice transactions still output V2, as earlier wallets won't know what to do with them anyhow)
- [x] Grin-wallet v1.1.0 will contain both the V1 and V2 Owner/Foreign APIs. The V2 APIs use the V2 slate by default.
- [x] SlateV2 in grin-wallet 1.1.0 will contain a `block_header_version` field that indicates what version of the Grin Protocol it is intended for.
- [x] Grin-wallet 1.1.0 will fill the `block_header_version` field as 1 (consensus)
- [x] Grin-wallet 1.1.0 will halt if it detects the current block is later than the hardfork block (hardcoded rule).
- [x] Grin-wallet 1.1.0 will produce bulletproofs in the older recovery format
- [x] Grin-wallet 1.1.0 will not attempt to post transactions to a node if the node reported height is >= the hardfork height (effectively covered by 'halt' point above)

Grin-wallet 2.0.0 will fill the `block_header_version` field of the slate as 2
Grin-wallet 2.0.0 will refuse to process slates with a `block_header_version` of 1.0
Grin-wallet 2.0.0 will not post transactions to a grin node reporting a current block height < HF height
Grin-wallet/Grin node v2.0.0 will have a node API tweak to ensure grin-wallet 1.1.0 cannot communicate with grin-node 2.0.0
Grin-wallet 2.0.0 will turn off the V1 owner/foreign APIs, and use the V2 APIs only
Grin-wallet 2.0.0 will drop support for SlateV0 and SlateV1, and default to using SlateV2 only.
Grin-wallet 2.0.0 will only produce bulletproofs in the newer recovery format

### DIFF
--- a/src/bin/cmd/wallet.rs
+++ b/src/bin/cmd/wallet.rs
@@ -44,6 +44,29 @@ pub fn wallet_command(wallet_args: &ArgMatches<'_>, config: GlobalWalletConfig) 
 	let wallet_config = config.members.unwrap().wallet;
 
 	let node_client = HTTPNodeClient::new(&wallet_config.check_node_api_http_addr, None);
+
+	// TODO: Very temporary code to obsolete grin wallet for the first hard fork
+	// All tx operations call get_chain_height as a first order of business,
+	// so this is the most non-intrusive place to put this
+	match node_client.clone().chain_height() {
+		Ok(h) => {
+			if h >= 262080 {
+				let err_str = "This version of grin-wallet is obsolete as of block 252080. Please download v2.0.0 from https://github.com/mimblewimble/grin-wallet/releases";
+				error!("{}", err_str);
+				println!();
+				println!("**************");
+				println!("{}", err_str);
+				println!("**************");
+				println!("(You can still view your balances by disconnecting from the grin node, however you will not be able to transact until you upgrade)");
+				println!();
+				return 1;
+			}
+		},
+		// just continue if can't connect to node, as user won't be able to do
+		// anything meaninful anyhow
+		Err(_) => {},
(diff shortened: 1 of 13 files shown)


src/bin/cmd/wallet.rs
```diff
@@ -44,6 +44,29 @@ pub fn wallet_command(wallet_args: &ArgMatches<'_>, config: GlobalWalletConfig)
 	let wallet_config = config.members.unwrap().wallet;
 
 	let node_client = HTTPNodeClient::new(&wallet_config.check_node_api_http_addr, None);
+
+	// TODO: Very temporary code to obsolete grin wallet for the first hard fork
+	// All tx operations call get_chain_height as a first order of business,
+	// so this is the most non-intrusive place to put this
+	match node_client.clone().chain_height() {
```
garyyu: There's no `node_api_secret` parameter on line 46 when creating a new node_client (`None` there), here the `chain_height()` calling will fail if node API need secret.

Issue #125, opened 2019-05-29
Recover without previously caling init fails


hashmap:
With grin-wallet 1.1 if I run `recover` without  previously calling `init` it fails with:

```
 ERROR grin_wallet_controller::command - Error recovering seed - IO error
```
I found that it expects `wallet_data` folder to exists, which is not the case in this scenario.

yeastplume:
Addressed #134

PR #128, opened 2019-05-31
fix: wallet initial recover from recovery phrase fail 

garyyu:
fix https://github.com/mimblewimble/grin-wallet/issues/125

(cherry picked from commit 61cdaf64e9cedac91167a395bec7b3912cc870e6)


yeastplume:
I need to look into this, but I think there was a reason why I chose not to allow creation of the wallet_data directory. Will look into after 1.1.0 release.

yeastplume:
Was addressed in #134

PR #127, opened 2019-05-31
filter feature for txs/outputs command

garyyu:
In production environment, when there're thousands of outputs and txs, the wallet command output will always show all the outputs or txs, which has too many outputs and not easy to read.

This PR give an enhancement to filter the display.

For `outputs`, we can query by
- a minimum output value
- and/or an `OutputStatus` (unconfirmed/unspent/locked/spent)
- and/or limit the total display lines

```
  - outputs:
      about: Raw wallet output info (list of outputs)
      args:
        - minvalue:
            help: Filter outputs by minimum value
            short: m
            long: minvalue
            takes_value: true
        - status:
            help: Filter outputs by status
            short: s
            long: status
            takes_value: true
        - limit:
            help: Limit the maximum display lines (show the latest ones)
            short: l
            long: limit
            takes_value: true
```
For example:
```
$ ./target/release/grin-wallet --floonet outputs -m 10.0 -s unspent -l 10
...
total displayed outputs: 10
```

For `txs`, we can query by
-  tx type (i.e. `coinbase/rx/tx/rxc/txc`  for coinbase/received/sent/ReceivedCanceled/SentCanceled)
- and/or tx creation date from and to
- and/or limit the total display lines

```
  - txs:
      about: Display transaction information
      args:
        - id:
            help: If specified, display transaction with given Id and all associated Inputs/Outputs
            short: i
            long: id
            takes_value: true
        - type:
            help: Filter txs by tx type (Coinbase/rx/tx/rxc/txc, means Received/Sent/ReceivedCanceled/SentCanceled)
            short: t
            long: type
            takes_value: true
        - startdate:
            help: Filter txs by creation date from 'startdate' to 'enddate', with format '1970-01-01 00:00:00'
            short: s
            long: startdate
            takes_value: true
        - enddate:
            help: Filter txs by creation date from 'startdate' to 'enddate', with format '1970-01-01 00:00:00'
            short: e
            long: enddate
            takes_value: true
        - limit:
            help: Limit the maximum display lines (show the latest ones)
            short: l
            long: limit
            takes_value: true
```
For example:
```
$ ./target/release/grin-wallet --floonet txs -l 50 -s "2019-05-02 00:00:00" -e "2019-05-04 00:00:00" -t tx

...

total displayed txs: 2
```

PR #132, opened 2019-06-02
Add pagination to retrieve_txs API call

invis87:


yeastplume:
Hi, thanks for the PR. Could you please provide a description of what the PR addresses and what it hopes to accomplish?

I'm not sure the notion of 'Pagination' is something that needs to be considered in the API. Why can't clients take care of it at the display level?

yeastplume:
There's also a suggestion for a more fully-featured filter mechanism here, which you might consider adding to or weighing in on somehow: https://github.com/mimblewimble/grin-wallet/pull/127

DavidBurkett:
@yeastplume Pagination is usually done on the server side for performance reasons. The retrieve_txs call is kind of slow, so it makes sense to do this here.

invis87:
@yeastplume Hi. This PR is for pagination on retrieve_tx API, issue for it: https://github.com/mimblewimble/grin/issues/2776
Pagination should be done on both sides to save traffic and improve speed.

yeastplume:
I can take that point. Have had more of a think, and I do think this functionality should be thought of in the context of a more generic query filter rather than bolting on two specific arguments to the API, such as in #2776 (where it's all done client side, so perhaps an amalgamation of approaches is needed). We can discuss this between now and post 2.0.0

yeastplume:
Closing due to lack of activity, feel free to bring up again in wallet subteam.

PR #154, opened 2019-06-15
Read version info from server, react accordingly

yeastplume:
Follow up from https://github.com/mimblewimble/grin/pull/2897 on the wallet side

* Change `v1/pool/push` endpoint call to `v1/pool/push_tx`
* Add function to read version info from node to NodeClient trait, add implementation to HttpNodeClient
* On startup, attempt to read version info from node and:
    * If can't communicate with node, continue read only
    * If node reports 404 for `v1/version` endpoint, assume node is < 2.0.0 and exit
    * If node reports version < what wallet is looking for as a minimum (currently 2.0.0-beta.1), exit with message to upgrade
    * ~~If node reports blockheader version < what wallet is looking for (currently set to 1 for testing, but should be bumped to 2 before releasing), exit with prompt to upgrade~~

### DIFF
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -16,7 +16,7 @@
 //! invocations) as needed.
 use crate::api::{self, ApiServer, BasicAuthMiddleware, ResponseFuture, Router, TLSConfig};
 use crate::keychain::Keychain;
-use crate::libwallet::{Error, ErrorKind, NodeClient, WalletBackend};
+use crate::libwallet::{Error, ErrorKind, NodeClient, WalletBackend, Slate, NodeVersionInfo};
 use crate::util::to_base64;
 use crate::util::Mutex;
 use failure::ResultExt;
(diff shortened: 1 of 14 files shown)


yeastplume:
* Added a pseudo-middleware function into the foreign API, which accepts a NodeVersionInfo struct and a slate and can return either Ok() or throw a libwallet error that will get propagated properly depending on how the API is instantiated (command line or json-rpc)
* Changed the get_version_info function for the node client to instead return an option. The http implementation will cache the version info value so it will only be called on the first invocation and the cached value used on subsequent calls
* Remove block header version check from statup as it can now go into middleware function

controller/src/controller.rs
```diff
@@ -39,6 +39,14 @@ lazy_static! {
 		HeaderValue::from_str("Basic realm=GrinOwnerAPI").unwrap();
 }
 
+fn check_middleware(
```
yeastplume: Middleware goes here (may have to redefine the middleware call to pass in the name of the function as well)

yeastplume:
Added middleware logic, which should work as follows:

* Pre HF, the wallet should build slates with the `version.block_header_version` field filled as 1. Post HF, the wallet will fill this field as 2.
* Pre HF, the wallet foreign_api will allow incoming slates with `version.block_header_version` set to 1 (and 1 only)
* Post HF, the wallet foreign_api will stop processing transactions where the incoming slates has its `version.block_header_version` set to < 2
* the foreign api's build_coinbase method does enforce any restrictions (other than the whole wallet needing to run against a 2.0.0 node) (build_coinbase can always be called, and will always be producing the correct proof according to block height)

PR #176, opened 2019-07-04
add --txid to the `wallet txs` command

### LINKED ISSUE #71: please add --txid to the `wallet txs` command
Currently the `wallet txs` command only accepts a transaction index number with `-i $index`, which makes things cumbersome. What we know is the txid, and in order to see the outputs, then right now I have to first list all transactions with `wallet txs` then parse the list to find the txid, then and fetch its $index, then run `wallet txs -i $index`. Adding --txid can remove the intermediary steps so I can see the outputs easily.

Could that be added please?

hendi:
This fixes #71

### DIFF
--- a/controller/src/command.rs
+++ b/controller/src/command.rs
@@ -587,6 +587,7 @@ pub fn outputs(
 /// Txs command args
 pub struct TxsArgs {
 	pub id: Option<u32>,
+	pub tx_slate_id: Option<Uuid>,
 }
 
 pub fn txs(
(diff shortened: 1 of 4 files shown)


yeastplume:
Looks great, thanks for doing this. I'm not sure why this was never exposed in the CLI as it's been in the API for some time.

Would you be able to add a quick test into `src/bin/cmd/wallet_tests.rs` that exercises this?

hendi:
Of course!

yeastplume:
Excellent, thanks again for this!

PR #180, opened 2019-07-06
Refactor - Split WalletCommAdapter into multiple traits

bddap:
This is WIP, consider it experimental for now.

Edit: Ready for merge


bddap:
Did a bit of investigation on this and it looks like this feature warrants a new WalletCommAdapter.

[WalletCommAdapter](https://github.com/mimblewimble/grin-wallet/blob/7b6d597d39a9fbc635fe6768c72370f80a2a05f8/impls/src/adapters/mod.rs#L30-L55) currently exposes a [https://deterministic.space/elegant-apis-in-rust.html#dont-stringly-type-your-api](stringly typed api) where the 'addr' argument is a string that represents a different type depending the Implementer.

I'm going to start by refactoring WalletCommAdapter, and refactoring implementors of WalletCommAdaptor to hold their address as struct fields. Not only will this make server authentication implementable, it will also be more strongly typed and will enable future efforts such as an [i2p](https://github.com/mimblewimble/grin/issues/2712) comm adaptor.

bddap:
WalletCommAdapter declares a [listen](https://github.com/mimblewimble/grin-wallet/blob/7b6d597d39a9fbc635fe6768c72370f80a2a05f8/impls/src/adapters/mod.rs#L47) method, but listen() is not implementable for all implementers of WalletCommAdapter. For example, it doesn't make sense for the FileWalletCommAdapter to listen(). Splitting the WalletCommAdapter trait in twain, or a similar solution, is next on my list after the task mentioned in the previous comment.

yeastplume:
Yes, that trait definitely needs a re-think.

Slightly related, I'm going to propose that we make the [grinbox](https://github.com/vault713/grinbox) spec a published open standard, and provide support for it in the default wallet. If you're able to look into that and have a think about it as part of all this work, I think it will be very valuable to help reduce the friction for the vast majority of transaction cases.

bddap:
Definitely agree. From what I can tell, grinbox comms can be implemented as another WalletCommAdapter. I'll keep grinbox support as an explicit end goal of these refactors.

impls/src/adapters/http.rs
```diff
@@ -144,32 +140,24 @@ impl WalletCommAdapter for HTTPWalletCommAdapter {
...
-		unimplemented!();
+impl Into<Error> for SchemeNotHttp {
+	fn into(self) -> Error {
+		let err_str = format!("url scheme must be http",);
+		ErrorKind::GenericError(err_str).into()
 	}
 }
 
-pub fn post<IN>(url: &str, api_secret: Option<String>, input: &IN) -> Result<String, api::Error>
+pub fn post<IN>(url: &Url, api_secret: Option<String>, input: &IN) -> Result<String, api::Error>
```
bddap: `&str` is changed to `&Url` in pursuit of [type safety](https://rust-lang-nursery.github.io/api-guidelines/type-safety.html).

impls/src/adapters/keybase.rs
```diff
@@ -280,33 +296,23 @@ fn poll(nseconds: u64, channel: &str) -> Option<Slate> {
...
 
-impl WalletCommAdapter for KeybaseWalletCommAdapter {
-	fn supports_sync(&self) -> bool {
-		true
-	}
-
-	// Send a slate to a keybase username then wait for a response for TTL seconds.
-	fn send_tx_sync(&self, addr: &str, slate: &Slate) -> Result<Slate, Error> {
-		// Limit only one recipient
-		if addr.matches(",").count() > 0 {
```
bddap: Check moved to constructor.

impls/src/adapters/mod.rs
```diff
@@ -15,41 +15,80 @@
...
+	fn send_tx(&self, slate: &Slate) -> Result<Slate, Error>;
+}
 
+pub trait SlateReceiver {
 	/// Start a listener, passing received messages to the wallet api directly
 	/// Takes a wallet config for now to avoid needing all sorts of awkward
 	/// type parameters on this trait
 	fn listen(
 		&self,
-		params: HashMap<String, String>,
```
bddap: Params was a) loosely typed, and b) unused.

impls/src/adapters/mod.rs
```diff
@@ -15,41 +15,80 @@
...
+}
+
+/// Checks for a transaction from a corresponding SlatePutter, returns the transaction if it exists
+pub trait SlateGetter {
+	/// Receive a transaction async. (Actually just read it from wherever and return the slate)
+	fn get_tx(&self) -> Result<Slate, Error>;
+}
+
+/// select a SlateSender based on method and dest fields from, e.g., SendArgs
+pub fn create_sender(method: &str, dest: &str) -> Result<Box<dyn SlateSender>, Error> {
```
bddap: If this method does not work in the context of I2P or public key based server authentication, it won't be much trouble to simply delete it. It's only used in two places.

impls/src/adapters/null.rs
```diff
@@ -1,59 +0,0 @@
-// Copyright 2018 The Grin Developers
```
bddap: A null adapter is no longer needed and is never used.

impls/src/test_framework/testclient.rs
```diff
@@ -344,55 +343,6 @@ impl LocalWalletClient {
 	}
 }
 
-impl WalletCommAdapter for LocalWalletClient {
```
bddap: This implementation was never used.

bddap:
So far none of these changes have nothing to do with Public Key based server authentication. Instead the changes so far are general improvements paving the way for new comm adapters.

I'm going to rename this pr, limiting scope to the refactor only.

bddap:
Requesting merge.

Key changes:
- Add typed fields to comm adapters to avoid passing of stringly typed arguments to WalletCommAdapter methods.
- Split WalletCommAdapter into four separate traits, one for each behavior it previously tried to encompass.

New traits:

- `SlateSender` - Sends transactions to a corresponding SlateReceiver
- `SlateReceiver` - Listens for slates from SlateSender
- `SlatePutter` - Asynchronous version of SlateSender; Posts slates to be read later by a corresponding SlateGetter
- `SlateGetter` - Asynchronous version of SlateReceiver; Checks for a transactions from a corresponding SlatePutter, returns the transaction if it exists

Implementation matrix:

|               | PathToSlate | HttpSlateSender | KeybaseAllChannels | KeybaseChannel |
| -             | -           | -               | -                  | -              |
| SlateSender   | ✗           | ✓               | ✗                  | ✓              |
| SlateReceiver | ✗           | ✗               | ✓                  | ✗              |
| SlatePutter   | ✓           | ✗               | ✗                  | ✗              |
| SlateGetter   | ✓           | ✗               | ✗                  | ✗              |

yeastplume:
All looks really good, thanks for doing this refactor and yes, the previous trait didn't make a lot of sense, this is much more coherent.

I'm going to merge this into the `milestone\2.1.0` branch (we're keeping master for any required fixes or changes to 2.0.0). This should give us plenty of time to test before the target release anyhow. Also need to merge these changes onto my own current refactor at https://github.com/mimblewimble/grin-wallet/pull/184, so we'll need a decent round of testing before 2.1.0 anyhow.

PR #184, opened 2019-07-09
Full Lifecycle API Support

yeastplume:
Exploratory PR being developed in tandem with: https://github.com/mimblewimble/grin-rfcs/pull/11

Thus far, this:

* Creates a new wallet Trait, WalletLCProvider, which is meant to provide implementations of the lifecycle methods as found in the RFC
* Redefines the 'WalletInst' trait to one that contains all the necessary components of a wallet, and use that trait everywhere where a wallet instance is held
* Put explicit lifetimes on the wallet traits, to avoid issues with `'static` lifetime requirements
* Default implementation of a WalletLCProvider (DefautLCProvider) and WalletInst (DefaultWalletInst) in the impls crate, which should replace the older static instantiation methods and allow for the wallet backend to be re-loaded and swapped in/out etc during a running instance

Also, this is an all-out war with the Rust compiler, but getting there.


yeastplume:
* Remove 'WalletConfig' from underlying wallet traits, backend implementations should only need the data directory
* change 'open_with_credentials' in WalletBackend to 'set_keychain', opening of the keychain and setting it in the wallet should be the job of a lifecycle provider.
* Once the wallet keychain is opened, it remains open until end of execution or 'close_wallet' is called. This is in contrast to the previous state, where it was re-opened during every API call.

yeastplume:
Ready for any review now, paves the way for changes in the associated RFC without actually changing anything external for users or wallet developers. 

Note there's an internal change whereby the seed is now kept in memory instead of the password as before when the wallet is listening. This won't be the case at the next release, where we'd be expecting to replace this with an XORed token value (see current proposal of security model in RFC)

api/src/foreign.rs
```diff
@@ -462,11 +471,23 @@ macro_rules! doctest_helper_setup_doc_env_foreign {
 			.unwrap();
 		let mut wallet_config = WalletConfig::default();
 		wallet_config.data_file_dir = dir.to_owned();
-		let pw = "";
+		let pw = ZeroingString::from("");
```
rentenmark: it would be cool to see a lazy static for this. C# for example has `string.Empty` etc

api/src/owner_rpc.rs
```diff
@@ -1338,7 +1340,8 @@ pub fn run_doctest_owner(
 ) -> Result<Option<serde_json::Value>, String> {
```
rentenmark: this function has a lot of `unwrap`. I understand that it is a test helper but even so, it seems un-necessary not sure how difficult the refactor of this return type would be

config/src/config.rs
```diff
@@ -109,28 +115,35 @@ fn check_api_secret_file(chain_type: &global::ChainTypes) -> Result<(), ConfigEr
...
 ) -> Result<GlobalWalletConfig, ConfigError> {
-	check_api_secret_file(chain_type)?;
+	check_api_secret_file(chain_type, data_path.clone())?;
 	// Use config file if current directory if it exists, .grin home otherwise
 	if let Some(p) = check_config_current_dir(WALLET_CONFIG_FILE_NAME) {
 		GlobalWalletConfig::new(p.to_str().unwrap())
 	} else {
 		// Check if grin dir exists
-		let grin_path = get_grin_path(chain_type)?;
+		let grin_path = match data_path {
```
rentenmark: could probably use `unwrap_or`, it's a style question

config/src/config.rs
```diff
@@ -109,28 +115,35 @@ fn check_api_secret_file(chain_type: &global::ChainTypes) -> Result<(), ConfigEr
...
+			Some(p) => p,
+			None => get_grin_path(chain_type)?,
+		};
 
 		// Get path to default config file
 		let mut config_path = grin_path.clone();
 		config_path.push(WALLET_CONFIG_FILE_NAME);
 
-		// Spit it out if it doesn't exist
+		println!("CONFIG PATH: {}", config_path.to_str().unwrap());
```
rentenmark: should this be `info!` ?

controller/src/controller.rs
```diff
@@ -131,35 +135,36 @@ where
 
 	// If so configured, add the foreign API to the same port
 	if owner_api_include_foreign.unwrap_or(false) {
-		info!("Starting HTTP Foreign API on Owner server at {}.", addr);
-		let foreign_api_handler_v2 = ForeignAPIHandlerV2::new(wallet.clone());
+		warn!("Starting HTTP Foreign API on Owner server at {}.", addr);
```
rentenmark: why `info`->`warn` ? if we want to see the messages we can set the log level, these don't look like warnings to me

controller/src/controller.rs
```diff
@@ -131,35 +135,36 @@ where
...
 		.join()
 		.map_err(|e| ErrorKind::GenericError(format!("API thread panicked :{:?}", e)).into())
 }
 
 /// Listener version, providing same API but listening for requests on a
 /// port and wrapping the calls
-pub fn foreign_listener<T: ?Sized, C, K>(
-	wallet: Arc<Mutex<T>>,
+pub fn foreign_listener<L, C, K>(
+	wallet: Arc<Mutex<Box<dyn WalletInst<'static, L, C, K> + 'static>>>,
```
rentenmark: this type is unfortunate but i understand how it came to be

api/src/foreign.rs
yeastplume: Think this is okay, it's just test setup

api/src/owner_rpc.rs
yeastplume: I think unwraps in tests are okay for now, happy to see a PR later that removes them (a lot of these tests were written before you could return a Result from main or from a test).

config/src/config.rs
yeastplume: Leftover debug println, thanks for catching!

controller/src/controller.rs
yeastplume: Changed this for consistency with how it's done in other places. The default in grin.toml is `warn`, and I think we want to make sure something like that is definitely printed to the console by default.

controller/src/controller.rs
yeastplume: Indeed... the joys of rust. At least it works how you think it's going to work once you're done fighting with the compiler

yeastplume:
Thanks for review, wallet APIs are definitely exercised extensively by tests, the CLI is as well but a bit less so. I've manually tested that all of the init/recover/restore operations work exactly as they did before these changes.

More tests are definitely needed, however I'm planning to replace most of the CLI code with the CLI code borrowed from wallet713, to make the whole thing more cohesive. Will definitely be adding to the test suite at that stage to ensure all operations are tested at the CLI/listener level as well.

yeastplume:
Yes, next bits of functionality to go in related the the current RFC are much more easily split-up. I don't think there will be a need for another refactor of this size (in the foreseeable future, anyhow).

yeastplume:
#212

PR #186, opened 2019-07-12
[WIP] Server authentication in transactions

bddap:
Adds a way to send grin to a server authenticated by its TLS public key. PKI is not required or used. The public key must already be known by the sender.


bddap:
Blocked by https://github.com/mimblewimble/grin-wallet/pull/180

src/bin/cmd/wallet_args.rs
```diff
@@ -621,18 +615,17 @@ pub fn parse_process_invoice_args(
...
-	// Now we need to prompt the user whether they want to do this,
-	// which requires reading the slate
-	#[cfg(not(test))]
-	let adapter = FileWalletCommAdapter::new();
-	#[cfg(not(test))]
-	let slate = match adapter.receive_tx_async(&tx_file) {
-		Ok(s) => s,
-		Err(e) => return Err(ParseError::ArgumentError(format!("{}", e))),
-	};
+	if cfg!(not(test)) {
```
bddap: When using `#[cfg(not(test))]`, tests were compiling and running successfully, but non-test builds were failing. This change makes sure type checks take place in both test mode and non-test mode.

bddap:
I won't have time to finish this one, at least not in the near future. If someone else could pick up where I left off it would be much appreciated.

Some background:

These changes make use of the new trait, `SlateSender`, introduced in https://github.com/mimblewimble/grin-wallet/pull/180.

The the struct this pr introduces is called `AuthenticatedHttpsSlateSender`. AuthenticatedHttpsSlateSender is meant to implement SlateSender.

The implementation is complete except for the public key verification bit. I've left one rustc error behind to indicate the main unfinished part of this pr. Just run `cargo check` and you'll see what I mean.

Work so far is on the client (SlateSender). The server shouldn't require any code; just run the foreign api on https and note the public key used.

Once AuthenticatedHttpsSlateSender is implemented, authenticated sends can be added as a method to the wallet's send command.


Feel free to ask me questions.

yeastplume:
Going to leave this open for now as I think it's a good idea, but it's likely going to need discussion with the wallet subteam and possibly an RFC. I'm busy with other things at the moment, but if anyone wants to pick this up now help would be welcome.

yeastplume:
Going to close for now, as the focus is on TOR at the moment which handles the encryption without needing to provide extra keys to senders. If somebody's interested in looking at this further, please feel free to propose via RFC.

PR #202, opened 2019-08-05
use VersionedSlate for the owner_rpc api

### LINKED ISSUE #201: Owner API should leverage VersionedSlate and not Slate directly
Currently the Owner API uses `Slate` directly. This is too closely coupled to the internals of grin core. We should break this coupling and leverage `VersionedSlate`.

Are there any reasons _not_ to do this or why this would cause problems?

antiochp:
Resolves #201 

Takes the same approach as the foreign_rpc api where we use `VersionedSlate` on the rpc side of things, converting to `Slate` internally.
These both serialize/deserialize in an identical fashion currently so this does not affect anything.
But once we start modifying `Slate` itself (via the contained `Transaction`) we should be decoupled from these changes.


yeastplume:
All fine, will merge on top of my other PR and then change the new set of 'secure' functions to use versioned slate as well.

PR #205, opened 2019-08-12
Test refactoring and additions to support API Testing

yeastplume:
In anticipation of adding the lifecycle API and security features into a V3 owner API, some test refactoring with the aim of:

* Developing proper end-to-end integration tests for the V3 API as it's developed (ensuring tests exist that are invoked via the HTTP listener in addition to the API doctests)
* Ensuring some end-to-end V2 owner API testing is in place to ensure backwards compatibility
* Insert some tests for the foreign API to ensure backwards compatibility between wallets/slate versions

Still WIP, will refine comments with more info as it develops


yeastplume:
#212

PR #206, opened 2019-08-13
Add `init_api_secure` function

yeastplume:
* Adds a new V3 Owner API, with the V2 API still running for backwards-compatibility purposes
* Adds functionality to the controller to encrypt and decrypt responses. All V3 API functions except `init_secure_api` must be encrypted.
* Secure requests are JSON-RPC calls that look like the following:
```
{
   "jsonrpc": "2.0",
   "method: "encrypted_request_v3",
   "id": "1",
   "params": {
      "nonce": "ef32...",
      "body_enc": "e0bcd..."
   }
}
```
* Encryption is currently AEAD With the following parameters:
    * AES-256 in GCM mode with 128-bit tags and 96 bit nonces 
    * 12 byte nonce which must be included in each request/response to use on the decrypting side
    * Empty vector for additional data
    * Suffix length = AES-256 GCM mode tag length = 16 bytes 
    * Open to discussion on the encryption scheme, but AEAD/AES256GCM is currently supported in node.js by default so hopefully this encryption shouldn't throw up too much of a block.
* Adds an `init_secure_api` function to the V3 API, performs an ECDH exchange (using the usual libsecp256k1 curve, again supported by node.js), sets the shared key internally and returns the public key to the caller.
* The shared-key on the server side is static and global (as I don't believe hyper method handlers can mutate their state)
* Also adds many tests exercising the V3 api encryption and error conditions


tests/owner_v3_init_secure.rs
```diff
@@ -0,0 +1,217 @@
...
+	// 10) Call 'init_secure_api' unencrypted (which we can do) and negotiate new shared key
+	let req = include_str!("data/v3_reqs/init_secure_api.req.json");
+	let res = send_request(1, "http://127.0.0.1:3420/v3/owner", req)?;
+	println!("RES 10: {:?}", res);
+
+	assert!(res.is_ok());
+	let value: ECDHPubkey = res.unwrap();
+	let shared_key = derive_ecdh_key(sec_key_str, &value.ecdh_pubkey);
+
+	// 12) A normal request, correct key
```
DavidBurkett: 11 :smile:

controller/src/controller.rs
```diff
@@ -308,15 +415,54 @@ where
 		OwnerAPIHandlerV3 { wallet }
 	}
 
+	//TODO: Unwraps
```
DavidBurkett: Has this been addressed?

tests/common/mod.rs
```diff
@@ -287,10 +319,86 @@ where
...
+	sec_req_id: u32,
+	internal_request_id: u32,
+	dest: &str,
+	req: &str,
+	shared_key: &SecretKey,
+) -> Result<Result<OUT, WalletAPIReturnError>, api::Error>
+where
+	OUT: DeserializeOwned,
+{
+	let url = Url::parse(dest).unwrap();
```
DavidBurkett: Should we really have all of these unwraps in `send_request_enc` and `derive_ecdh_key`? What RPC error code would be returned if these panic?

yeastplume: It's just test helper code, I don't usually worry too much about unwraps there. I've been careful to ensure the server side returns proper errors for all conditions (and never unwraps)

controller/src/controller.rs
yeastplume: Yes, leftover reminder comment.

PR #208, opened 2019-08-16
Fix typos in subcommand descriptions

phyro:
I did not test this locally. I did grep the project for the typos and it's the only place I've found them. You're free to close this if it's not acceptable without testing it


yeastplume:
Thanks for pointing this out! Slipped changes to the text into: https://github.com/mimblewimble/grin-wallet/pull/211

PR #129, opened 2019-05-31
rework kernel features support

antiochp:
This PR allows grin-wallet to build successfully against grin given the updated code in https://github.com/mimblewimble/grin/pull/2859

The `fee` and `lock_height` were moved into the relevant kernel feature variants for type safety.
This PR reworks how we handle kernel features to support this.

Depends on: https://github.com/mimblewimble/grin/pull/2859

Introduce a `KernelFeaturesCompat` enum to allow "versioned slate" json serialization/deserialization to be unaffected by these changes. 

----

Cargo.toml references grin libs via "bleeding edge" aka master here so will not successfully build until we merge the changes in https://github.com/mimblewimble/grin/pull/2859 to master.

PR #209, opened 2019-08-19
fix for base url when checking other version via foreign api

antiochp:
This resolves what looks like a bad merge at some point on master branch, most likely during post-HF branch cleanup.

The current `2.0.x` branch has the correct impl of this - 

https://github.com/mimblewimble/grin-wallet/blob/7b6d597d39a9fbc635fe6768c72370f80a2a05f8/impls/src/adapters/http.rs#L105-L108

At some point this was broken or reverted or missed on master and we are not using the correct url when hitting the foreign api to "check current version".

These changes are included in https://github.com/mimblewimble/grin-wallet/pull/206 but I'd like to get them in independently to help with testing for https://github.com/mimblewimble/grin-wallet/pull/129 "Rework kernel features support".

### DIFF
--- a/impls/src/adapters/http.rs
+++ b/impls/src/adapters/http.rs
@@ -36,7 +36,7 @@ impl HttpSlateSender {
 	}
 
 	/// Check version of the listening wallet
-	fn check_other_version(&self) -> Result<(), Error> {
+	fn check_other_version(&self, url: &Url) -> Result<(), Error> {
 		let req = json!({
 			"jsonrpc": "2.0",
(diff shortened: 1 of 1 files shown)


yeastplume:
I've merged #206 so should be good, test failures are oddball circle-ci windows specific things, so I've addressed them in that PR as well. Sorry for rebase needs, but I hope it should be fairly straightforward, won't start anything else until this work is all in and tested

antiochp:
Closing - #206 includes this.

PR #210, opened 2019-08-20
Introduce VersionedCoinbase to add api stability around build_coinbase

antiochp:
This PR introduces a `VersionedCoinbase`, following the same approach as the existing `VersionedSlate`.

We have a `build_coinbase` api call on the foreign api that is used by a mining node to generate a new coinbase reward (output + kernel) when building a candidate block for mining.

It is important this api continues to work even if we make internal changes to the core output and kernel impls.

The `build_coinbase` api endpoint now returns a `VersionedCoinbase` (currently uses `V2`) and this maintains full backward compatibility with the existing node and wallet.

This api endpoint is a little tricky and easy to miss because it works in the opposite direction to everything else. This is accessed `node->wallet` when generating candidate blocks for mining.

This PR should add some additional robustness to ensure old nodes can mine using a newer wallet and vice-versa.

Related - #129 

TODO - 

- [x] local testing of mining nodes and wallets to confirm we can still mine successfully in various old/new combinations...

### DIFF
--- a/api/src/foreign_rpc.rs
+++ b/api/src/foreign_rpc.rs
@@ -17,7 +17,8 @@
 use crate::keychain::Keychain;
 use crate::libwallet::{
 	self, BlockFees, CbData, ErrorKind, InitTxArgs, IssueInvoiceTxArgs, NodeClient,
-	NodeVersionInfo, Slate, VersionInfo, VersionedSlate, WalletLCProvider,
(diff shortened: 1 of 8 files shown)


yeastplume:
LGTM, and looks like it should work. We need to be sure we've done adequate testing against older wallets (and particularly a few exchanges before the next official release.

PR #216, opened 2019-09-06
Update transaction status based on kernel lookup

jaspervdm:
WIP


jaspervdm:
Yeastplume is taking over development of this feature, so closing this PR.

PR #220, opened 2019-09-16
Update transactions via kernel where necessary

yeastplume:
Implementation of kernel lookups, current thinking is to keep the current output-based assumptions in place, track cases where a sender has no change outputs and only perform kernel-lookups to see whether a transaction has been confirmed in that instance.

Still more testing, edge cases and test implementations required, but what's there at the moment seems to work (see the `no_change` test in `controller\tests`


jaspervdm:
Is this PR ready for review?

yeastplume:
Yes, please by all means

libwallet/src/internal/selection.rs
```diff
@@ -134,6 +140,11 @@ where
 		let filename = format!("{}.grintx", slate_id);
 		t.stored_tx = Some(filename);
 		t.fee = Some(slate.fee);
+		// TODO: Future multi-kernel considerations
+		if total_change == 0 {
```
jaspervdm: What is the reason for only doing the kernel lookup when there is no change? I think it is a more robust method overall than by change output. If for example someone is sending with `-c 0` or is using the same seed on multiple wallets at the same time, the tx could stay in unconfirmed state forever, even though it was included in a block.

yeastplume: I was of 2 minds about this, in the vast majority of cases the output state should be enough to infer the transaction state (and output update operations are still happening whenever the wallet is queried,) so it seemed to make sense to limit kernel lookups where they aren't necessary and only burden the server with them in cases where it's evident the lookup is needed. The kernel excess needs to be stored in the transaction entry in order for the wallet to query for it afterwards, so in the case where someone's using the same seed on multiple wallets at the same time, the 'other' wallet doesn't know the excess or the existence of the transaction anyhow (and its output state is very inconsistent with the truth so a 'check' will be needed, at which point the transaction log will be inferred from the outputs anyhow).

I think we need more thought here anyhow, we have plans to look at the entire update process, (and most likely include the 'check' as part of normal operations.) So perhaps we can leave this as is for now just to solve the immediate bug, and then undertake to try and find the optimal balance of kernel lookups/output lookups/scan operations that keep the wallet state as consistent as possible in all cases, even between 2 wallets using the same seed and in 0 conf situations, etc?

jaspervdm: > and output update operations are still happening whenever the wallet is queried

True, thats why i mentioned the `-c 0` case.

> so in the case where someone's using the same seed on multiple wallets at the same time, the 'other' wallet doesn't know the excess or the existence of the transaction anyhow

Sure, but if the change output i spent on another wallet with the same seed before the tx is marked as confirmed, it will stay unconfirmed forever.

However, I do agree that both those cases are really edge situations, and in most realistic scenarios this PR is sufficient.

> ..and then undertake to try and find the optimal balance of kernel lookups/output lookups/scan operations that keep the wallet state as consistent as possible in all cases

Agreed, lets merge this and take some time to rethink the wallet state update mechanism as a whole.

PR #225, opened 2019-09-26
use explicit protocol version (local is not safe)

antiochp:
This is a bit tricky.

A 2.1.0 node uses protocol version 2 internally via `ProtocolVersion::local()`.
But for backward compatibility expects txs via the push api (from the wallet) in proto version 1.

Building grin-wallet pulls in `ser` from node repo.
So we cannot use `ProtocolVersion::local()` in the wallet repo.

This only surfaced as an issue once we tagged node and wallet repos and started building against these.


antiochp:
Going to merge this now - we need to explicitly define `ProtocolVersion(1)` like this for 2.1.0 wallet to interact successfully with 2.1.0 node.

Issue #229, opened 2019-10-01
Always store kernel excess in db


lehnberg:
_As proposed by @jaspervdm during dev meeting. Full discussion on gitter: https://gitter.im/grin_community/dev?at=5d9373c9eb1eff63d65a27bf_

---

### Description
* Currently, the kernel excess of a tx is stored only if the user is the sender and there is no change output. 
* This should change so that the kernel excess is always stored for every transaction inserted into the db, regardless if the user is a sender or receiver.

### Motivation
* It can't be calculated after the fact, so if the user needs it sometime later for whatever reason, they are out of luck;
* Users can use it to quickly find a particular tx on block explorers;
* Exchanges/pools/payment providers can use it to check on-chain status without relying on the grin-wallet. They can also communicate it to their users as something that is analogous to an on-chain txid

### Note
* Not every tx will always have a excess associated to it (for example it won't be recoverable when restoring from seed).
* While introducing this change, it MUST be ensured that the code that's selectively doing the kernel lookup is not triggered. It's currently triggered based on the presence of a stored kernel excess.

PR #231, opened 2019-10-02
Always store and update kernel excess

yeastplume:
Addresses #229 

* Always store and update the kernel excess when a transaction is created
* Store kernel excess when receiving funds
* Recalculate and store excess when confirming coinbase outputs
* Only lookup by kernel if the transaction's `amount_credited` or `amount_debited` is zero
* Use new Option<Commitment> serializer to output kernel excess, modify doctests
* Output kernel excess on command line during `grin-wallet txs` call

Relies on https://github.com/mimblewimble/grin/pull/3074


yeastplume:
Good call, updated and top comment updated. Also had to update serializers, and recalculate the excess when confirming coinbase outputs, if you could just check over then I think we're good to here.

PR #236, opened 2019-10-09
Fix URL Compatibility with 2.0.0

yeastplume:
Unintended consequence of earlier changes to how urls are handled meant that everything but the base URL was being stripped when sending to another wallet. This fix reverts to earlier behaviour, keeping the entire URL+Path intact.

Going to merge into master first, then cherry-pick into the `current/2.1.x` branch


yeastplume:
Removed URL lib altogether (which is opinionated about how URLs should be formed). 

Stuff like the following should be allowed (as in 2.0.0) 

`https://poloniex.com/public?currency=GRIN&command=createDeposit&id=2a35f07d-1e55-4cfe-86a0-6660ec2cc6c5`

So reverted back to just simple appending of strings to form the final destination url.

Issue #244, opened 2019-10-17
Rework of wallet output and transaction state update process


yeastplume:
Placeholder issue for definition of how wallets should update themselves from a node. We should take the opportunity to finalize 'best practices' here, and define a process that keeps the wallet and related transactions in sync as well as possible.

Right now, the process is roughly (will update with more detail as needed):

### Coinbase
   * Create a potential coinbase output for a mined block
   * If the block is accepted the output is detected on the chain, create a transaction log entry of type 'Confirmed Coinbase', with 'confirmed' set to true

### Transacting (payer -> payee)
   * The sender creates a new 'blank' transaction Slate, adding inputs and change outputs to the slate.
   * The sender generates a random kernel offset
   * The sender sends the slate to the payee (via file or http). If sending synchronously, (e.g via http)  the associated transaction is saved to the log after a response from the payee's listener. If sending asynchronously, (e.g. via file), the transaction is saved immediately. 
   * When saved, at this stage, the associated transaction is set to type 'Sent Tx'  with a status of 'unconfirmed', inputs are locked internally and a change output is added with status 'unconfirmed'.
   * The payee receives the slate, creates (an) output(s) for the received amount with status unconfirmed, and immediately stores a transaction in their log of type 'Received Tx' with confirmed set to false. 
   * The payee calculates and saves the transaction kernel for later reference.
   * The slate is returned to the payer for completion, who calculates and saves the transaction kernel.

(Invoicing is similar to the above, with the roles of payer/payee and sender/receiver swapped, so I won't detail it separately here)

### Updating Transaction/Output State
   * On every info/send/txs command, query all 'unconfirmed' outputs in the user's wallet against the node.
   * If an 'unconfirmed' output in the user's wallet appears on the chain, change the status to 'confirmed', and update the associated transaction status to 'confirmed'.
   * If the transaction has no change outputs and the wallet is the payer, also perform a kernel lookup when performing a 'txs' command.

### Check-Repair

Check-Repair currently fixes most of the potential issues that come up via the current updating model, namely:
* Removing transactions that have never confirmed, setting their state to cancelled, unlocking and/or deleting associated outputs as needed.
* Adding outputs that exist on the chain but for some reason or another are not on the chain, creating transactions for them
* Removing outputs that may exist as 'Unspent' but don't exist on the chain (via a fork, for instance)
* Note this process is identical to the restore process

At the moment, this is a manual process that needs to be initiated by the user.

yeastplume:
[...]

PR #246, opened 2019-10-21
Store and use last scanned PMMR height in `check_repair`

yeastplume:
As part of #244, it will be necessary to store the last scanned height during a `check_repair`, and part of the wallet update process should be to 'check_repair` back a threshold of blocks from the last scanned height. This PR aims to add that functionality, as well as modifying the existing check_repair method to accept a number of block heights back to start scanning from.

May require some changes or an addition to the node API to support this properly, still exploring what the best thing to do is here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -683,17 +683,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "grin_api"
 version = "3.0.0-alpha.1"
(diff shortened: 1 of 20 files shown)


yeastplume:
This is in a working state, so would like to merge work on this to date shortly as part of works for https://github.com/yeastplume/grin-rfcs/blob/master/text/0000-wallet-update-process.md. In the current state, the incremental scanning by block height is complete and working. 

Would like to avoid a single mega-PR for this work, so subsequent PRs will fill in some of the details details according to how the RFC develops.

PR #249, opened 2019-10-31
Tx combinator refactor

### LINKED ISSUE #248: refactor tx building combinators
Grin node "transaction building combinators" are changing here -  https://github.com/mimblewimble/grin/pull/3057

We use some of these directly in grin-wallet when building basic transactions interactively.
These changes will require some (relatively minor) changes to grin-wallet to reflect these.

Example of building a tx via the new explicit tx kernel feature passing, replacing the existing `with_fee()` combinator - 

```
let tx = transaction(
    KernelFeatures::Plain { fee: 4 },
    vec![
        input(6, key_id1),
        output(2, key_id2),
    ],
    &keychain,
    &builder,
)
```

Related - https://github.com/mimblewimble/grin/pull/3057

antiochp:
Resolves #248 
Related https://github.com/mimblewimble/grin/pull/3057 (must be merged before this PR)

We no longer use tx building combinators for fee and lock_height.
We now pass an existing tx into `build::partial_transaction`.
We are also responsible for updating the tx kernel to reflect the features based on the current slate (fee changes based on the tx being built).

The motivation for these changes is to increase the flexibility to support future kernel feature variants, specifically "relative height locked" kernels.
The tx building combinators `with_fee` and `with_lock_height` did not fit cleanly with adding new feature variants as the different combinations of optional fields (fee, lock_height etc.) produce different kernel features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -683,7 +683,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "grin_api"
 version = "3.0.0-alpha.1"
-source = "git+https://github.com/mimblewimble/grin#da2e75299191acd70d4a48bd63c3dcb2cfcc74d2"
+source = "git+https://github.com/mimblewimble/grin#50ce7ba0433e916ddf6154ff64bdd987bd0e68c2"
 dependencies = [
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


antiochp:
Just merged the related grin PR and kicked off a build here.

antiochp:
Bumped grin dependency in Cargo.lock to reflect the changes on master.

antiochp:
We should keep an eye on tx building using master for a bit - should be pretty obvious if this broke something unexpected.

PR #252, opened 2019-11-04
Change check_repair + certain functions to lock more granularly

yeastplume:
To support changes in https://github.com/mimblewimble/grin-rfcs/pull/30, this PR aims to:

* Rename `check_repair` to `scan`
* Modify the internal `scan` implementation to lock wallet contents a more granular level, freeing up the ability for other threads to run against the wallet while a scan is running.
* Modify internal versions of functions that rely on `scan` to also lock at a more granular level.
* Remove the `restore` function in favor of the single `scan` function.
* Automatically perform a scan when a wallet is created by recovering from seed
* Modify the command line `scan` command to optionally accept a start block height for the scan


yeastplume:
This is ready for review now if anyone wants to take a look

PR #254, opened 2019-11-12
unnecessary trailing semicolon

jcomito:
warning: unnecessary trailing semicolon
   --> impls/src/backends/lmdb.rs:365:81


yeastplume:
thanks for pointing this out, slipped the fix into another PR

PR #255, opened 2019-11-15
Add option to finalize and post transactions separately (for air-gap wallets)

phreaknik:
# What does this PR add?
This PR adds the ability to finalize a transaction without posting the transaction to the chain, and separately post a finalized slate to the chain.

# Why is this useful?
1) **For air-gapped wallets**: Currently there is no easy way to send transactions from an offline/air-gapped wallet. To finalize a transaction, the wallet requires a connection to a running node. With this feature, it is now possible to generate a finalized slate on an offline machine (without posting to chain), copy that slate file to another machine (with a connected Grin node), and post the transaction on that machine.

2) **For sender privacy**: Currently, the sender must reveal his IP to the network (or at least to one connected node) during transaction finalization. With this feature, a transaction sender no longer needs to reveal his/her IP to the network. The sender may simply generate a finalized transaction slate, and give that slate to the recipient, allowing the recipient to reveal his/her IP instead.

# Testing
I have functionally tested this change on Mainnet. I tested the new segregated finalize/post transaction method, as well as the original combined finalize+post method. Both transaction methods work as expected.

I am unclear about the test process for this project. I am happy to write formal test cases, if someone can give me a bit of guidance on where/how those tests should be added.

### DIFF
--- a/controller/src/command.rs
+++ b/controller/src/command.rs
@@ -406,6 +406,8 @@ where
 pub struct FinalizeArgs {
 	pub input: String,
 	pub fluff: bool,
+	pub nopost: bool,
+	pub dest: Option<String>,
 }
 
 pub fn finalize<'a, L, C, K>(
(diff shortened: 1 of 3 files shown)


yeastplume:
This looks good, and thanks for the PR. I think this has been requested before, and it's a nice optional feature add.

casey:
Thank you @phreaknik, super useful!

PR #259, opened 2019-11-20
Payment Proof Implementation

yeastplume:
Aims to add payment proofs as described in https://github.com/mimblewimble/grin-rfcs/pull/31 to the wallet APIs.

Changes to the Wallet OwnerAPI is as follows:

* Addition of `payment_proof_recipient_address` to `InitTxArgs`. If present, this will add the sender address and recipient address to the slate.
* Addition of 2 helper functions:
   * `get_public_proof_address`, which returns the wallet's proof address at the given derivation path (which corresponds to the onionv3 address generated for TOR)
   * `proof_address_from_onion_v3` - pure helper that extracts the proof address from an onionv3 address

If the `payment_proof` is included in the slate, the recipient will fill out the signature as outlined in the RFC.

Note that for now there is only 1 payment address per wallet account at derivation path 0. It's likely this will be enhanced in the future. The address provided will generally need to match the address at derivation path 0 of the listening wallet.

Also note there's no assumption here that TOR is involved, so the init TX function needs to accept an address without caring where it's from. Integrating this into the TOR workflow and adding command line functionality that integrates the functionality of this PR into TOR will be the focus of the next PR.


yeastplume:
Top comment updated with details and ready for review.

jaspervdm:
Looks good to me! So checking if the payment proof is actually present when it is mandatory (in the case of TOR) would not happen at the wallet level, but at the transport level right?

yeastplume:
Great, thanks for reviewing. The payment proof isn't mandatory at any time, but when it's requested (by including the receiver address in the slate), the sender should always check the result and reject the transaction if the proof hasn't been filled out correctly. All happens in the lower layers, so callers shouldn't have to do anything extra.

PR #260, opened 2019-11-27
Proof of Payment Command Line

yeastplume:
Requires #259 (this PR will look much chunkier than it is until that's merged) and aims to:
* Add proof of payment commands to the command line, thus far we have
   * `request_payment_proof` and `proof_address` added to the `send` command. `proof_address` is only required if the destination isn't an onionV3 address
* Display payment proof information for the `txs` command if only viewing a single transaction (so to see proof info `grin-wallet txs -i 2`
* Enable TOR sending in the API via the `init_send_tx` command (currently disabled for backwards compatibility reasons, no longer a factor.
* Add a display address command so users without TOR can obtain their wallet-derived public address, the private key of which is used for signing payment proofs.


yeastplume:
Ready for review and merge (after #259 is reviewed and merged!)

PR #261, opened 2019-11-28
TTL Implementation

yeastplume:
This provides the implementation of TTL functionality within the wallet API. Note that this PR doesn't yet add the TTL to the command line client or add any default values (which likely need a bit of further discussion).

Changes Are:

* TTL Field as it appears in the slate is now an Option<u64>, meant to specify the block height at which a transaction should not be further processed.
* All functions that process a transaction, `recieve_tx`, `finalize` etc now do a check for the TTL value before continuing processing. Note the wallet uses the stored last confirmed height for this value as opposed to calling out to the server to get the value (so as not to create a dependency on a node running in order to transact)
* The wallet update process now cancels all transactions for which the TTL has expired (this check is actually done against a running node)
* `InitTxParams` now has a `ttl_blocks` field specifying the number of blocks from current to set slate's TTL.


controller/src/controller.rs
```diff
@@ -64,8 +64,8 @@ fn check_middleware(
 			}
 			if let Some(s) = slate {
 				if s.version_info.version < CURRENT_SLATE_VERSION
-					|| (bhv == 2 && s.version_info.block_header_version != 2)
-					|| (bhv > 2 && s.version_info.block_header_version < GRIN_BLOCK_HEADER_VERSION)
+					|| (bhv == 3 && s.version_info.block_header_version != 3)
```
DavidBurkett: Since txns from block header 2 are compatible with those from block header version 3, you can just remove line 67. We only needed that logic because of the bulletproof serialization changes.

yeastplume: :+1:

PR #268, opened 2019-12-02
Remove manual recovery

yeastplume:
This changes the `recover` command to display the recovery phrase only. A much more intuitive and non confusing recovery can be done with `init -r` now, which will automatically restore via the scan command on first run after recovery.


libwallet/src/slate.rs
```diff
@@ -187,8 +187,8 @@ pub struct Slate {
 	/// TTL, the block height at which wallets
 	/// should refuse to process the transaction and unlock all
 	/// associated outputs
-	#[serde(with = "secp_ser::string_or_u64")]
-	pub ttl_cutoff_height: u64,
+	#[serde(with = "secp_ser::opt_string_or_u64")]
+	pub ttl_cutoff_height: Option<u64>,
```
quentinlesceller: Seems unrelated but probably fine though https://github.com/mimblewimble/grin-wallet/pull/267 ?

PR #272, opened 2019-12-03
V2 Slate Compatibility

yeastplume:
This restores wallet-to-wallet compatibility between V3.0.0 and V2.0.0 by re-introducing the V2 slate. The rules for conversion are:

When sending via HTTP (including TOR), the slate version on the remote side is checked and:
   * If V3 support is reported back, send slate as a V3 slate
   * If V2 only is reported back:
      * Downgrade slate to V2 before sending if `payment_proof` is None
      * If `payment_proof` is Some then stop the transaction with an error saying proofs aren't supported by the other side
      * If `ttl_cutoff_height` is some, warn the user that it will be ignored by the other side and dropped by them in the slate, but continue to downgrade to V2

When outputting a file (either via send or invoice):
   * If both `payment_proof` and `ttl_cutoff_height` are None, downgrade slate to V2 before outputting
   * If either `payment_proof` or `ttl_cutoff_height` are Some, output V3 slate with a warning that the other side must be upgraded to grin-wallet v3.0.0 or later

Tested sending back and forth between v2.0.0 and 3.0.0 wallets, sends via direct http and files, also invoices.

### DIFF
--- a/api/src/foreign_rpc.rs
+++ b/api/src/foreign_rpc.rs
@@ -553,9 +553,10 @@ where
 		message: Option<String>,
 	) -> Result<VersionedSlate, ErrorKind> {
 		let version = in_slate.version();
+		let slate_from = Slate::from(in_slate);
 		let out_slate = Foreign::receive_tx(
 			self,
-			&Slate::from(in_slate),
+			&slate_from,
 			dest_acct_name.as_ref().map(String::as_str),
 			message,
 		)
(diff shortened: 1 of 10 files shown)


quentinlesceller:
Currently reviewing but just to let you know that the failing tests is because of:
```
thread 'main' panicked at '(left != right) 
left: {
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "Ok": {
      "foreign_api_version": 2,
      "supported_slate_versions": [
        "V3",
        "V2"
      ]
    }
  }
}
right: {
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "Ok": {
      "foreign_api_version": 2,
      "supported_slate_versions": [
        "V3"
      ]
    }
  }
}'
```
Defined here https://github.com/mimblewimble/grin-wallet/blob/master/api/src/foreign_rpc.rs#L56 
So you don't have to go on Azure website :).

PR #284, opened 2020-01-01
Fix: Log update thread do not exit after call ownerAPI.

jdwldnqi837:
Log update thread do not exit after call ownerAPI.


yeastplume:
I'm just looking at this now, and it seems like this change would allow the thread to exit when the sender channel goes out of scope (as opposed to the parent thread going out of scope,) but I'm left guessing as to what the intent is. It would help if you could submit a clearer, more detailed description of what the symptoms are for the issue that this fixes and why you believe this change is needed.

yeastplume:
I've looked at this and the suggested behaviour is correct (the loop in the log thread exits otherwise). Going to merge this into master but also cherry-pick to 3.0.1. Thanks for pointing this out!

PR #295, opened 2020-01-14
Wallet Interactive-mode, Part 1

yeastplume:
The goal of the PR will be to have all of the command line options working in CLI mode at a basic level, perhaps not all polished or perfect with respect to their functionality. 

It's important to note that this adds command line functionality while leaving the existing single-invocation mode functionality intact. The CLI functionality can be introduced, then refined and tweaked over time without any impact on existing functionality.

The focuses of this PR are:

* Provide a basic command-prompt mode for all existing wallet actions without duplicating code or functionality.
* Ensure the commands and switches are consistent between the CLI and single-invocation mode (except for commands that only make sense in one or the other)
* Ensure the CLI mode is backed by a single static instance of the Owner/Foreign APIs, to ensure they operate as similarly to upstream wallets as possible.
* Enable whatever additional automated testing of the Owner/Foreign APIs the new CLI mode affords that couldn't have taken place before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -974,6 +974,7 @@ dependencies = [
  "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.44 (registry+https://github.com/rust-lang/crates.io-index)",
+ "termion 1.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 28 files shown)


yeastplume:
Still quite janky, but basically works. Going to merge this shortly as experimental functionality to allow users to start giving feedback and do further refinements over time.

For now, cli mode is totally optional and its presence has no effect on existing functionality.  To run, it's simply `grin-wallet cli`. Once cli mode is considered complete enough, it can be run as the default when `grin-wallet` is run without args.

Issue #306, opened 2020-01-21
Use V3 Onion address format exclusively in all user-facing contexts


yeastplume:
In the 3.0.0 release, there is a bit of inconsistency whereby some commands accept a wallet's address as a raw public key whereas others use the v3 Onion address format. In retrospect we should only be presenting one format to users, not both.

For the 3.1.0 release, we should try and use the v3 Onion address exclusively wherever we can, which will include
* When specifying payment address manually on the command line, the user should be able to input an onion v3 address, not a raw key
* When retrieving addresses for the wallet, the user should only be shown the ov3 address
* When showing/exporting payment proof only the ov3 address should be used.

The slate format should not be changed for this, only user-facing instances where the address is displayed

The JSON-RPC APIs should be modified to accept either/or in init_send_tx via a custom formatter.

yeastplume:
Done via #309

PR #309, opened 2020-01-22
Onion V3 address consistency

yeastplume:
Implementation of onionv3 address cleanup as described in https://github.com/mimblewimble/grin-wallet/issues/306.

Also refactors the existing OnionV3 Address conversion code into a proper type, to clean up conversion code a bit and make future development that much cleaner.


yeastplume:
Tested and ready for merge

PR #314, opened 2020-01-22
Secure API JSON RPC ID can be integer or string

### LINKED ISSUE #312: API v3: ID is sometimes int or string
The RPC id is sometimes a string and sometimes an integer.
Example:
`init_secure_api`
```
  "id": "1",
  "jsonrpc": "2.0",
  "result": {
    "Ok": "03ef8e5c18ff365039fa0292281e70e2881aacdbac9818e7941be3af890fd9e028"
  }
}
```

broken `open_wallet` query
```
{
  "error": {
    "code": -32002,
    "message": "Encrypted request format error: invalid type: string \"2\", expected u32"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

### LINKED ISSUE #313: API v3: Secure API does not accept string ID
Similar to #312 except a bit more annoying this time.
Once initialized, the secure API does not accept ID as string.
```
{ 
   "id":"3",
   "jsonrpc":"2.0",
   "method":"encrypted_request_v3",
   "params":{ 
      "nonce":"2994319f73303015805b22c3",
      "body_enc":"M1X8zvf0MTjJ8uYkEOWlxKcRvz16g+PtiKNdttYddlMapOzGi8gV7yJrsOYdDkaLhVP3Y3o8LTGuVW56EpV+s8Tjgxy9DWoPIiW1vMm5V0ITvXX2enZVgswmkvNjnkhjTB1tydpv"
   }
}
{
  "error": {
    "code": -32002,
    "message": "Encrypted request format error: invalid type: string \"3\", expected u32"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

quentinlesceller:
Fixes #312 and #313.
uint64 are now serialized as string and can be deserialized either from integer or string.

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -142,14 +142,15 @@ pub struct EncryptedRequest {
 	/// method
 	pub method: String,
 	/// id
-	pub id: u32,
+	#[serde(with = "secp_ser::string_or_u64")]
+	pub id: u64,
 	/// Body params, which includes nonce and encrypted request
 	pub params: EncryptedBody,
 }
 
 impl EncryptedRequest {
 	/// from json
-	pub fn from_json(id: u32, json_in: &Value, enc_key: &SecretKey) -> Result<Self, Error> {
+	pub fn from_json(id: u64, json_in: &Value, enc_key: &SecretKey) -> Result<Self, Error> {
 		Ok(EncryptedRequest {
 			jsonrpc: "2.0".to_owned(),
 			method: "encrypted_request_v3".to_owned(),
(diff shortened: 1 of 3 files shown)


yeastplume:
Looks good (small test complication failure), but I'm wondering if it's more correct to leave the id as a u32 and just provide a 'string_or_u32' formatter as well. JS usually only supports 32 bit literals (which is why all u64s throughout the API are returned and accepted as strings).

quentinlesceller:
True, that's one way to do it. But since easyjson-rpc https://github.com/mimblewimble/easy-jsonrpc/blob/master/src/lib.rs#L468 uses u64 internally, wouldn't it be logical to also use u64?

yeastplume:
Yeah, I don't think it makes a huge difference. Thanks for fixing!

Issue #279, opened 2019-12-16
Error wallet.seed and grin-wallet.toml exists when restoring from seed


hackinthebochs:
When attempting to restore a wallet from mneumonic seed, the application first requests the seed, then requests a wallet password. Only after these are successfully entered does it check for the pre-existence of the wallet.seed and grin-wallet.toml files and exits with an error if they already exist. 

If you typed the wallet seed manually, this can be extremely frustrating to have to type it all again for an error the wallet could have checked for before the data was entered. This issue is also compounded by the fact that the file existence checks happen one at a time, which may require multiple attempts at creating the wallet including re-typing in the seed and wallet password.

yeastplume:
Okay, valid point and worth looking into

yeastplume:
Fixed via #318

PR #318, opened 2020-01-29
Rework directory checks/setups at launch

### LINKED ISSUE #308: top_level_dir continues to broken
--top_level_dir continues to be broken and has not really worked since I reported an issue when it was named data_dir in the last versions of grin.

**What is the Issue?**
--top_level_dir does NOT load the grin-wallet.toml that was defined in the custom directory. Grin will correctly make the grin-wallet.toml in that directory if it does not exist. It will even set all the paths correctly for the api secrets and logging in said generated grin-wallet.toml.

However, it does NOT load the configuration file and utilize it as part of it's configuration and therefore it is impossible to run with a custom config path.

**Reproducing The Issue**
This can be verified running these two commands (notice the ".grin2" and not ".grin"):

1) grin -t /home/user/.grin2/ init
2) grin -t /home/user/.grin2/ info

When (1) is run it will create a wallet_data and grin-wallet.toml file as expected in /home/user/.grin2/. If you look at the grin-wallet.toml file you will see all config options have correctly set paths to /home/user/.grin2/ for everything from the api_secret to the wallet logging.

However, you will notice when running (1) or (2) that grin continues to ignore /home/user/.grin2/grin-wallet.toml. Instead, it will create files in /home/user/.grin/. Running the wallet command will fill the incorrect log path /home/user/.grin/main/grin-wallet.log with information.

This should ALL be put inside of /home/user/.grin2/ and the log should be at /home/user/.grin2/grin-wallet.log. This is where it is defined to be at via the "log_file_path" that grin auto-created when it auto-greated the grin-wallet.toml.

yeastplume:
Reworks initial directory setup logic and the flow of the `init` command to:

* That the `--top-level-dir` switch is respected appropriately
* That the api secret files aren't written to default locations when using the `--top-level-dir` switch or initting a wallet in the current directory (these were erroneously being created in the `~./grin/main` dirs)
* Check for the presence of an existing configuration file *before* asking user for a password/recovery phrase, and exit if so.

Fixes #308 and #279

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -42,7 +42,10 @@ pub const API_SECRET_FILE_NAME: &'static str = ".api_secret";
 /// Owner API secret
 pub const OWNER_API_SECRET_FILE_NAME: &'static str = ".owner_api_secret";
 
-fn get_grin_path(chain_type: &global::ChainTypes) -> Result<PathBuf, ConfigError> {
+fn get_grin_path(
+	chain_type: &global::ChainTypes,
+	create_path: bool,
+) -> Result<PathBuf, ConfigError> {
 	// Check if grin dir exists
 	let mut grin_path = match dirs::home_dir() {
 		Some(p) => p,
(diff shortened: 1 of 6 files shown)


config/src/config.rs
```diff
@@ -119,34 +136,52 @@ fn check_api_secret_file(
...
+				(grin_path, default_config)
+			}
+			true => {
+				let mut path = config_path.clone();
+				path.pop();
+				(
+					path,
+					GlobalWalletConfig::new(config_path.to_str().unwrap())?,
+				)
+			}
```
quentinlesceller: Considering using `if/else` instead of `match` here
```
		// Return defaults if file doesn't exist
		if config_path.clone().exists() {
			let mut default_config = GlobalWalletConfig::for_chain(chain_type);
			default_config.config_file_path = Some(config_path);
			// update paths relative to current dir
			default_config.update_paths(&grin_path);
			(grin_path, default_config)
		} else {
			let mut path = config_path.clone();
			path.pop();
			(
				path,
				GlobalWalletConfig::new(config_path.to_str().unwrap())?,
			)
		}
```

yeastplume: I kind of prefer match syntax even for simple cases like this, it makes the false and true conditions more explicit, and harder to mix them up when maintaining (like you may have done in your example) (:rofl: Sorry!)

yeastplume:
Excellent, thanks very much for reviewing and testing.

config/src/config.rs
quentinlesceller: I understand 😆

PR #319, opened 2020-01-30
Ensure API `create_wallet` returns failure when provided with invalid mnemonic seeds

### LINKED ISSUE #292: owner_api v3 "create_wallet" return "ok" even receive invalid mnemonic seeds.
Pass invalid seeds to owner_api v3 "create_wallet" still got ""result": {"Ok": null}" response.
Although wallet.seed does not created in fact.

maybe add validate mnemonic code in this line 
https://github.com/mimblewimble/grin-wallet/blob/45480392da42afee448fdabbfd4a29591be6c5ab/impls/src/lifecycle/default.rs#L186
?

yeastplume:
Fixes #292, call to `WalletSeed::init_file` was swallowing its error and continuing.

Adds explicit tests to exercise `create_wallet` with both valid and invalid seeds.

### DIFF
--- a/api/src/owner_rpc_s.rs
+++ b/api/src/owner_rpc_s.rs
@@ -1517,7 +1517,7 @@ pub trait OwnerRpcS {
 		"params": {
 			"name": null,
 			"mnemonic": null,
-			"mnemonic_length": 0,
+			"mnemonic_length": 32,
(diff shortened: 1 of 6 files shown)


yeastplume:
thanks!

Issue #305, opened 2020-01-21
Switch node client to use Node v2 API


yeastplume:
Swap the existing HTTP node client to use the Node's new JSON-RPC API instead of the V1 API. Should be relatively straightforward as the functionality is all in one file, but this will need to be tested thoroughly as none of the automated tests use the HTTP node client.

quentinlesceller:
I can take this one :).

yeastplume:
@quentinlesceller nice!

PR #329, opened 2020-02-10
Convert node client to use V2 Node API

yeastplume:
Work in progress, includes new JSON-RPC client library code.

Will address #305

### DIFF
--- a/impls/src/client_utils/client.rs
+++ b/impls/src/client_utils/client.rs
@@ -147,7 +147,7 @@ impl Client {
 	/// object as body on a given URL that returns a JSON object. Handles request
 	/// building, JSON serialization and deserialization, and response code
 	/// checking.
-	pub fn _post<IN, OUT>(
+	pub fn post<IN, OUT>(
 		&self,
 		url: &str,
 		api_secret: Option<String>,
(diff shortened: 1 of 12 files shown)


impls/src/client_utils/client.rs
```diff
@@ -165,7 +165,7 @@ impl Client {
 	/// provided JSON object as body on a given URL that returns a future. Handles
 	/// request building, JSON serialization and deserialization, and response code
 	/// checking.
-	pub fn _post_async<IN, OUT>(
+	pub fn post_async<IN, OUT>(
```
quentinlesceller: Out of curiosity why not using `pub async fn`?

jaspervdm: This code is originally from the [grin repo](https://github.com/mimblewimble/grin/blob/6e5afe496b201a1a5df4f42c62cd5fe1bdc24c6b/api/src/client.rs).
[My PR there](https://github.com/mimblewimble/grin/pull/3214/files#diff-1ba0421725cb0e5e93523cb69ded4a46) is updating the code to use `pub async fn`, although iirc the wallet only uses the blocking function. Using the blocking functions many times is not really optimal since it spawns a full runtime for each request, but its something we can improve in a separate PR. For this PR i think its fine to leave as is

yeastplume: Indeed, focus is elsewhere on this PR, can likely make improvements later.

PR #332, opened 2020-02-13
CI: Fix macOS build

quentinlesceller:


quentinlesceller:
Actually I'm not sure it fix it in all the case.

PR #336, opened 2020-02-18
Change EncryptedRequest 'id' field to String

yeastplume:
(Replaces #334, with a saner commit history)

In doing some sample code for the V3 API, it became evident that the ID field in encrypted requests shouldn't be restricted to being a u64, or a string representation of a u64. It should allow all types of IDs in so long as they're strings, including UUIDs.

This changes the EncryptedRequest/Response fields to type String, and provides a custom legacy serializer to ensure anyone currently sending ids as ints won't be affected by the change.


yeastplume:
Tested locally, tests are consistently failing due to croaring.

PR #337, opened 2020-02-18
Cargo update

yeastplume:
* Update with latest upstream Grin dependencies in Cargo.lock
* libtx functions now require a SwitchCommitment as opposed to a reference.


yeastplume:
Merging, and will address constant croaring the failures separately.

PR #338, opened 2020-02-18
Remove Crate Linefeed

tenthousandlakesmn:
grin-wallet throws `Inappropriate ioctl for device (os error 25)` when `grin-wallet init -r` or `grin-wallet pay` is executed as a child process in node. It seems the crate linefeed is the culprit. 

Instead of spawning a separate terminal using linefeed, data is read from std::io.


yeastplume:
Hi @tenthousandlakesmn And thanks for submitting this PR. 

Unfortunately, I don't think removing the linefeed crate is the right approach. Linefeed was chosen deliberately over just using stdin for usability reasons; I'll just give a quick example why:

This is output from the current grin-wallet master when inputting a recovery phrase:

```
[yeastplume@yeastbox restore_338]$ ../grin-wallet-cur --floonet init -hr
Please enter your recovery phrase:
phrase> foot kiss barely husband absorb sheriff adapt obverse gadget air history retreat hour skill field swing lawsuit search cream dove host bike blush front

Recovery word phrase is invalid.

Please enter your recovery phrase:
phrase> foot kiss barely husband absorb sheriff adapt observe gadget air history retreat hour skill field swing lawsuit search cream dove host bike blush front
Please provide a new password for the recovered wallet
Password: 
```
I made a mistake entering the line above, and `obverse` above should have been `observe`. Linefeed presented the incorrect line back to me, and I was able to go back with arrow keys correct it, and have the phrase accepted.

Now the version using stdout:

```
Please enter your recovery phrase:
foot kis^[[A barely husband absorb sherrif adapt ob^[[Bserve gadget air history retreat hour skill field switng lawsuit search cream dove host bike blush frint

Recovery word phrase is invalid.
Please enter a valid recovery word phrase or Q to quit

^[[A
```
I've entered the phrase incorrectly, any attempt to use arrow keys are interpreted literally as character codes, and there's nothing buffering my input to present it back to me, so when I get it wrong, I have to type everything in again exactly as it was. Since I'm likely typing a phrase written on paper, this going to be a very frustrating experience.

So I don't think we can just remove linefeed unless this is addressed. Can I ask what your usecase is here? There may be other ways of addressing the underlying issue. Even if there's no other way around it, we could possibly include a command line option somewhere that optionallly reads input from stdin if it's toggled, but leaves the default functionality in place.

tenthousandlakesmn:
Hello @yeastplume :grin: 

Thank you for your feedback :rocket: 

I'm developing a wallet UI using node.js

I'm executing grin-wallet with the code: 

``` javascript
const config = require(`../config`);
const log = require(`../log`)(`wallet`);
const spawn = require(`child_process`).spawn;

let recoverProcess = spawn(`${config.walletPath}`, [`init`, `-r`], {
  shell: true,
  windowsHide: true
});

recoverProcess.stdout.on(`data`, (data) => {
  log.debug(`recover: ${data.toString()}`);
});
```

Which works for most commands.

What you're saying makes sense, this is more of an XY problem. 

I'm using the CLI because that's the only way to get that information from grin-wallet.

Ultimately, having RPC **recover** / **init** / **address** endpoints will make developing a wallet more straight forward.

I'm gonna focus on that :ice_hockey:

yeastplume:
No problem. Just so you're aware, there's already an API in place to get this information. It's documented in the code [or on crates.io here](https://docs.rs/grin_wallet_api/3.0.0/grin_wallet_api/trait.OwnerRpcS.html)

I've only just recently added a node.js sample outlining how to run it, which you may also find useful: https://github.com/mimblewimble/grin-wallet/tree/master/doc/samples/v3_api_node

PR #340, opened 2020-02-21
get_onion_address_v3 (lacks documentation)

tenthousandlakesmn:
This is an endpoint to get a onion address with the method `get_onion_v3_address`.

How does the `doctest_helper_json_rpc_owner_assert_response!()` macro work?

### DIFF
--- a/api/src/owner_rpc_s.rs
+++ b/api/src/owner_rpc_s.rs
@@ -30,6 +30,7 @@ use crate::util::secp::key::{PublicKey, SecretKey};
 use crate::util::{static_secp_instance, ZeroingString};
 use crate::{ECDHPubkey, Owner, PubAddress, Token};
 use easy_jsonrpc_mw;
+use grin_wallet_util::OnionV3Address;
 use rand::thread_rng;
 use std::time::Duration;
 
@@ -1826,6 +1827,15 @@ pub trait OwnerRpcS {
 	) -> Result<PubAddress, ErrorKind>;
 
 	/**
+	Needs documentation!
+	*/
+	fn get_onion_v3_address(
+		&self,
+		token: Token,
+		derivation_index: u32,
+	) -> Result<OnionV3Address, ErrorKind>;
+
+	/**
 	Networked version of [Owner::proof_address_from_onion_v3](struct.Owner.html#method.proof_address_from_onion_v3).
 	```
 	# grin_wallet_api::doctest_helper_json_rpc_owner_assert_response!(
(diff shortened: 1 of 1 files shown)


yeastplume:
Thanks again for PRing, appreciate it. I'll just give some background here for the benefit of everyone and then we can decide how to proceed.

The payment proof functionality came towards the end of development for the 3.0.0 HF release, and some of the details probably didn't get as much scrutiny as they should have due to all of the other things going on. The basic problem is I ended up accepting and displaying the V3 Onion address and the Raw Public Key address in different places, even though they represent the same thing. So the send command took a pubkey for the payment proof address (when sending to non-TOR addresses), and an OV3 address for the actual send address. It also ended up displaying both when you query transactions via the 'txs -id n` command.

I've since changed most of the output and anywhere there is user interaction to use OV3 addresses exclusively, and convert where necessary. Unfortunately the API is a bit muddled now, cause we have:

`get_public_proof_address`, which returns a raw address pubkey
`proof_address_from_onion_v3`, which is a helper to convert OV3 addresses to raw pubkeys.

Which are both now a bit broken now because all we really want is for the `get_public_proof_address` to return an OV3 key, which is accepted everywhere an address is needed, both via command line and the API.

Rather than introducing more functions, I think we actually just want to change `get_public_proof_address` to return an OV3 address, which can be provided by users and everywhere in the API where an address is needed. However, we want to get there while causing minimal disruption to existing users of the API. 

I think (though I'm not 100% certain,) that this should be okay for most users at the moment, but it's probably safer to do as you suggest, add a new function and mark the others as deprecated for 4.0. So a few thoughts here:

* I'd just rename what you have to `get_public_address` and mark the other 2 functions as deprecated to be removed in 4.0
* There shouldn't be any logic in the api wrapper functions, they should always be calling the actual OwnerAPI, which should expose the exact same functionality to developers who build directly. That's also where the functions are properly documented, see here for an example: https://github.com/mimblewimble/grin-wallet/blob/4f0c5c015fa6e08e7984a6febbda0e50028336fe/api/src/owner.rs#L1981
* The function as written won't serialize the OnionV3 return type properly, as it needs a wrapper that avoids creating extra structure, see here for an example: https://github.com/mimblewimble/grin-wallet/blob/4f0c5c015fa6e08e7984a6febbda0e50028336fe/api/src/types.rs#L51 
* The Doctest documentation should be filled out and passing. This is just filling out a json request and expected response. These will be run as automated tests in a simulated wallet environment to ensure they're working and that future changes don't break them. There are plenty of examples in the file, and the docs for similar functions should serve as a starting point, see: https://github.com/mimblewimble/grin-wallet/blob/4f0c5c015fa6e08e7984a6febbda0e50028336fe/api/src/owner_rpc_s.rs#L1796

Thanks, and let me know what you think!

PR #345, opened 2020-02-24
Fix CI

quentinlesceller:
Similar to https://github.com/mimblewimble/grin/pull/3233.


quentinlesceller:
Merging this to fix the CI (simply install `libclang` and `libclang-dev`)

Issue #317, opened 2020-01-27
Tx-building: Compact slates


lehnberg:
### Description

This issue tracks the effort to redesign slate structure into a new version that minimises the data shared in each trip and tries to reduce slate footprint as much as possible.

### Old description
> As per Round 2.1 in the [transaction building workflow](https://github.com/mimblewimble/grin-wallet/blob/master/doc/transaction/basic-transaction-wf.png), sender shares `change_outputs` information with recipient.
>
> This leads to the recipient gaining information of sender's future outputs that can be used to de-anonymize other outputs of the sender. Example: Alice makes a grin deposit to an exchange with KYC requirements, and as part of this creates a change output. This change output is later used as part of spending other outputs that the exchange has never seen. If the exchange monitors the chain or mempool for the particular output being spent, they can begin making assumptions about other outputs that are being spent in the same context. If an exchange sees heavy grin usage, they might be able to deanonymize a great deal of output commitments this way. 
>
>It would therefore improve privacy if change outputs were not shared between transacting parties.  Are there any immediate negative consequences, beyond the receiver not being able to verify that `transaction_fee` corresponds to the number of outputs being created?

---

_Edit:_ As per discussion in [grincoin#dev](https://keybase.io/team/grincoin):
* A receiver would still be able to learn the change output by monitoring the mempool, assuming tx is not aggregated when it is added there.
* One approach would be for sender not to send inputs or change output, just the kernel, with the receiver trusting that the input exists.
* If this information is passed along to the receiver needlessly, in addition to leaking data too easily, it can also be said to waste bandwidth.

_Edit Apr 02:_ Renaming issue, changing description, adding to 4.0.0 milestone.

yeastplume:
I'm okay with this functionality in general, a couple of points

* This only works in the single payer->payee->payer workflow, not for the case of multiple parties transacting and not for the invoice workflow. Whoever posts the transaction obviously needs to have a complete awareness of all the outputs involved.
* Slate could be amended to either provide sum or outputs themselves, think that's fine.
* Does this need a short RFC?

quentinlesceller:
Considering that the compact slate RFC has been merged =>https://github.com/mimblewimble/grin-rfcs/pull/49
And that the corresponding code have been also merged => https://github.com/mimblewimble/grin-wallet/pull/404

I'm closing this issue. Awesome work @yeastplume ! 🎆

PR #356, opened 2020-03-06
Make Transaction Field in V4 Slate Optional

yeastplume:
In anticipation of experimentation related to https://github.com/mimblewimble/grin-wallet/issues/317, this: 

* Changes the `tx` field in the V4 Slate to be an `Option<Transaction>` as opposed to a `Transaction`
* Adds a `tx_or_error()` function to the Slate, and use that in all places where the transaction was read from the slate directly. This will help ensure that the transaction is always populated where it's expected to be populated (i.e. everywhere, currently,) Any future cases in which Slates can be used in a 'reduced' form can be handled specially.

This shouldn't have any effect on the format of serialized slates, and from an external perspective, nothing should change other than slates being able to de/serialize without the Transaction field present. (Should add a test for this)

### DIFF
--- a/api/src/foreign.rs
+++ b/api/src/foreign.rs
@@ -441,13 +441,13 @@ where
 	/// // If result okay, send to payer, who will apply the transaction via their
 	/// // owner API, then send back the slate
 	/// // ...
-	/// # let mut slate = Slate::blank(2);
+	/// # let slate = Slate::blank(2);
 	///
-	/// let slate = api_foreign.finalize_invoice_tx(&mut slate);
+	/// let slate = api_foreign.finalize_invoice_tx(&slate);
 	/// // if okay, then post via the owner API
 	/// ```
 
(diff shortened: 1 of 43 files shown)


api/src/foreign.rs
```diff
@@ -441,13 +441,13 @@ where
 	/// // If result okay, send to payer, who will apply the transaction via their
 	/// // owner API, then send back the slate
 	/// // ...
-	/// # let slate = Slate::blank(2);
+	/// # let mut slate = Slate::blank(2);
```
jaspervdm: Maybe we should have 2 methods instead of 1: `.tx()` and `.tx_mut()` to have a distinction between regular and mutable borrow. This would prevent having to add `mut`s and `&mut` in places where we don't really need them

yeastplume: Yes, think that's fair. I was of two minds over this but there's probably a few too many things needed to be made mutable as a result of what's there currently

Issue #352, opened 2020-03-03
 HTTP Send: Invalid HTTP version parsed


quentinlesceller:
Grin-wallet v3.1.0.
Getting this error (and also #349 which might be unrelated) when sending http payouts. Tested with Poloniex, Kucoin and Hotbit.
No errror with grin-wallet v3.0.0.
 
```
20200303 20:08:10.453 ERROR grin_wallet_impls::adapters::http - Performing version check (is recipient listening?): Request error: Cannot make request: invalid HTTP version parsed
```

jaspervdm:
Seems like `hyper-timeout` does not play nice with the connector from `hyper_rustls`. I've filed an issue there: https://github.com/hjr3/hyper-timeout/issues/13

jaspervdm:
Fixed in v3.1.1

Issue #349, opened 2020-02-28
ERROR grin_wallet_impls::adapters::http - Performing version check (is recipient listening?): Request error: Cannot make request: channel closed


yangchunhe:
version:grin-wallet 3.1.0-beta.2
when I send coin by http,the error:
./grin-wallet send -d https://abc.xx.com/grin/v1/111 1

20200228 13:39:48.577 ERROR grin_wallet_impls::adapters::http - Performing version check (is recipient listening?): Request error: Cannot make request: operation was canceled: received unexpected message from connection
Wallet command failed: LibWallet Error: Client Callback Error: Performing version check (is recipient listening?): Request error: Cannot make request: operation was canceled: received unexpected message from connection

jaspervdm:
Fixed in v3.1.1

PR #357, opened 2020-03-07
Update hyper-timeout

jaspervdm:
Fixes https://github.com/mimblewimble/grin-wallet/issues/349
Fixes https://github.com/mimblewimble/grin-wallet/issues/352


yeastplume:
Thanks for looking after this, should be cherry-picked into master as well.

PR #398, opened 2020-04-30
v3.x.x - v4.0.0 wallet compatibility fixes

yeastplume:
This PR should cover whatever changes are necessary to ensure communication with V3 wallets continues to work for the period between the release of v4.0.0 and the date of HF3. Note that quite a bit of this will be temporary code, to be removed shortly after HF3.

### DIFF
--- a/api/src/foreign_rpc.rs
+++ b/api/src/foreign_rpc.rs
@@ -180,6 +180,8 @@ pub trait ForeignRpc {
 		&self,
 		slate: VersionedSlate,
 		dest_acct_name: Option<String>,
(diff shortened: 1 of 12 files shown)


yeastplume:
Update: all compatibility with wallet v3.x.x. working, these are the rules:

* http/tor sends will downgrade automatically via the version check with the other party
* Creating file transactions will output as V3 slate by default until the HF3 block, and then V4 afterwards. You can force output of a V4 slate with the temporary `--v4` argument for the `send` and `invoice` commands.
* `recieve` or `pay` will output the same version used for the input

PR #404, opened 2020-05-11
Compact slate merge

yeastplume:
Implementation of Compact Slates as defined in the [Compact Slates RFC](https://github.com/mimblewimble/grin-rfcs/pull/49)

Functionality should be exactly as defined in the RFC. In addition, this version of the code has been tested against 3.0.0 wallets and should be completely compatible with them.

There is a lot to review here, and very limited time before 4.0.0, so I'd ask the review focus to be:

* Ensuring there are no obvious security issues
* Ensuring what's here matches the RFC

Rust style issues, minor things that could be coded more efficiently, etc can be marked down, but I only intend on addressing any major issues that may come up for the time being.

The PR history making up this branch is basically all PRs from #366 Onwards. It should be much easier to review this series of PRs rather than looking at this PR's diff.

This will be merged along with the RFC at the end of the RFC's FCP, (May 21st, 2020). Again, time is very limited before 4.0.0 so I'd ask everyone to try and at least give this a once-over between now and then.


yeastplume:
@j01tz many thanks for reviewing, and yes, we'll aim to perform a more thorough review of everything for 4.0.0

PR #407, opened 2020-05-13
Enhanced offset creation

yeastplume:
Previously:

* A single offset is chosen by whoever initiates the transaction. Their part of the excess is calculated as a combination of their inputs and outputs with the offset subtracted. The rest of the transaction continues with no more regard for the offset.

Now:

* The offset is set to 0 at the beginning of the Transaction
* Each party chooses their part of the offset before signing round 1 (which is when their contribution to the excess is calculated), and commits to ONLY their outputs minus their part of the offset as their excess (i.e, their inputs are not included in this total). Each new offset contribution is added to the total offset on the slate before being passed to the next party.
* When inputs are added to the transaction, their blinding factors are subtracted from the offset to create a new offset.

With this change to how offsets and excesses are handled, we should be a stone's throw away from [mostly lock-free transactions](https://forum.grin.mw/t/mostly-lock-free-transactions/731) without actually having them implemented. This introduces the offset workflow while keeping the existing output locks in place, which should allow us to introduce lock-free transactions post 4.0.0 without breaking any compatibility since the offset/excess assumptions on either side of a transaction will be the same regardless of whether the outputs are pre-locked or chosen just before posting.

Note that quite a bit of the code is littered with `is_compact` conditionals since we still need to support 3.0.0 compatibility for a short period between release and HF3.


yeastplume:
Ready for review with all tests passing and compatibility tested with 3.0.0 wallets.

libwallet/src/internal/selection.rs
```diff
@@ -55,6 +55,10 @@ where
 	C: NodeClient + 'a,
 	K: Keychain + 'a,
 {
+	//TODO: Revise HF3. If we're sending V4 slates, only include
+	// change outputs in excess sum
+	let include_inputs_in_sum = if slate.is_compact() { false } else { true };
```
tromp, line 60: This is an eye sore to me:-(
 Please change to

let include_inputs_in_sum = !slate.is_compact();

yeastplume: Truly a classic. No idea where that evolved from.

libwallet/src/internal/selection.rs
```diff
@@ -65,20 +69,25 @@ where
...
 	// Update the fee on the slate so we account for this when building the tx.
 	slate.fee = fee;
 
-	let blinding = slate.add_transaction_elements(keychain, &ProofBuilder::new(keychain), elems)?;
+	let blinding = match elems.len() {
+		// in the event of no change, keep our blinding factor at zero.
+		// it will be adjusted when we add inputs
+		0 => BlindingFactor::zero(),
+		_ => slate.add_transaction_elements(keychain, &ProofBuilder::new(keychain), elems)?,
+	};
```
tromp, line 83: Why can't we pass empty elems to add_transaction_elements() ? (which I would also expect to yield BlindingFactor::zero())

libwallet/src/internal/selection.rs
```diff
@@ -496,11 +513,13 @@ where
...
-			parts.push(build::coinbase_input(coin.value, coin.key_id.clone()));
-		} else {
-			parts.push(build::input(coin.value, coin.key_id.clone()));
+	if include_inputs_in_sum {
+		for coin in coins {
+			if coin.is_coinbase {
+				parts.push(build::coinbase_input(coin.value, coin.key_id.clone()));
+			} else {
+				parts.push(build::input(coin.value, coin.key_id.clone()));
+			}
```
tromp, line 522: This code suggests  a method build::input(value, key_id, is_coinbase) to replace the current tow ?!

libwallet/src/slate.rs
```diff
@@ -332,7 +331,12 @@ impl Slate {
 		K: Keychain,
 	{
 		// Whoever does this first generates the offset
-		if self.participant_data.len() == 0 {
+		// TODO: Remove HF3
+		if self.participant_data.len() == 0 && !self.is_compact() {
+			self.generate_offset(keychain, sec_key, use_test_rng)?;
+		}
+		// Always choose my part of the offset, and subtract from my excess
+		if self.is_compact() {
```
tromp, line 339: Avoid repeating tests by
if !self.is_compact() {
  if self.participant_data.len() == 0 {
  }
} else {
}

Btw, is there no method self.participant_data.is_empty() ? comparing len() with 0 looks ugly...

libwallet/src/slate.rs
```diff
@@ -514,7 +518,7 @@ impl Slate {
 		// Generate a random kernel offset here
 		// and subtract it from the blind_sum so we create
 		// the aggsig context with the "split" key
-		self.tx_or_err_mut()?.offset = match use_test_rng {
+		let my_offset = match use_test_rng {
```
tromp, line 521: can't this be written as an if then else ? or doesn't that work at expression level?

libwallet/src/slate.rs
```diff
@@ -525,12 +529,26 @@ impl Slate {
...
 		};
 
-		let blind_offset = keychain.blind_sum(
+		if self.is_compact() {
+			let total_offset = keychain.blind_sum(
+				&BlindSum::new()
+					.add_blinding_factor(self.offset.clone())
+					.add_blinding_factor(my_offset.clone()),
+			)?;
+			self.offset = total_offset;
```
tromp, line 538: is there no more compact way to express self.offset += my_offset ?

libwallet/src/slate.rs
```diff
@@ -1088,8 +1106,8 @@ pub fn tx_from_slate_v4(slate: &SlateV4) -> Option<Transaction> {
 			}),
 		}
 	}
-	if slate.offset != BlindingFactor::zero() {
-		tx.offset = slate.offset.clone()
+	if slate.off != BlindingFactor::zero() {
+		tx.offset = slate.off.clone()
```
tromp, line 1110: What's the rationale for testing against zero? Is tx.offset already set to a nonzero value if slate.off == zero?

libwallet/src/slate_versions/v4.rs
```diff
@@ -34,7 +34,8 @@
 //! *  The `payment_proof` struct is renamed to `proof`
 //! * The feat_args struct is added, which may be populated for non-Plain kernels
 //! * `proof` may be omitted from the slate if it is None (null),
-//! * `offset` is added, which may be optionally included during S3 and I3 to ensure the transaction can be re-built entirely from the slate information. Used for delayed transaction posting.
+//! * `off` (offset) is added, and will be modified by every participant in the transaction with a random
+//! value - the value of their inputs
```
tromp, line 38: not the value of their inputs, but their blinding factor.

tromp:
I didn't notice anything wrong, but then I only looked at the surface level, lacking any detailed understanding of the code. My review then is that the changes all look plausible.

libwallet/src/internal/selection.rs
yeastplume: Yes, I can move the check into `add_transaction_elements`, but the fundamental issue is the `build` module (contained in grin core) throws the error if elems is empty. The `build` library is some of the the oldest untouched part of the code and could perhaps do with another revision at some stage.

libwallet/src/internal/selection.rs
yeastplume: Similar to above, this is in the core tx `build` lib and could stand some revision (but after HF3)

libwallet/src/slate.rs
yeastplume: The logic is split here to allow for quick and easy deletion of 3.0.0 compatibility code once HF3 is behind us. There's a lot of this sprinkled around the code, it will be gone shortly after 4.0.0 is released.

libwallet/src/slate.rs
yeastplume: Stylistic choice, I tend to prefer match statements over if-else (particularly when the block is returning a value) because I find there's less tendency for them to get confused once logic starts expanding

libwallet/src/slate.rs
yeastplume: No in the current code, but yes it's possible to implement operator traits to support this syntax.  Again this is up in old portions of the grin Keychain code so a revision to the library could be done at a later date.

libwallet/src/slate.rs
yeastplume: Left over from an earlier iteration, thanks for spotting.

libwallet/src/slate_versions/v4.rs
yeastplume: corrected

libwallet/src/internal/selection.rs
antiochp: Surely this should be -

```
if !slate.is_compact() { true } else { false };
```

yeastplume: No, inputs aren't included in the participant's excess calculation at this stage if the slate is v4 (read `is_compact` == `is_v4). The input excesses are subtracted from the offset later in the process.

yeastplume:
> I think this looks ok to me.
> The pre/post HF3 logic definitely complicates things though.

Indeed it does, but I've tried to mark off and isolate any bits of code that can be removed after HF3, so hopefully this will become a lot cleaner.

PR #413, opened 2020-05-22
Slatepack Pt. 3 - The Packening (SlatepackAddress implementation)

yeastplume:
* Adds the `SlatepackAddress` struct as specified in the slatepack RFC with all required conversion and serialization functions.
* Replaces x25519 keys in the slatepack struct with `SlatepackAddress`
* Refactors Slatepack functions into libwallet in anticipation of adding slatepack functions to the API.

### DIFF
--- a/libwallet/src/slatepack/address.rs
+++ b/libwallet/src/slatepack/address.rs
@@ -101,7 +101,7 @@ impl From<&SlatepackAddress> for OnionV3Address {
 impl TryFrom<&SlatepackAddress> for xDalekPublicKey {
 	type Error = Error;
 	fn try_from(addr: &SlatepackAddress) -> Result<Self, Self::Error> {
-		println!("Slatepack ed25519 key: {:?}", addr.pub_key);
+		//println!("Slatepack ed25519 key: {:?}", addr.pub_key);
 		let cep =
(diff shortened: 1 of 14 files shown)


libwallet/src/slatepack/address.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+impl From<&SlatepackAddress> for OnionV3Address {
+	fn from(addr: &SlatepackAddress) -> Self {
+		OnionV3Address::from_bytes(addr.pub_key.to_bytes())
+	}
+}
+
+impl TryFrom<&SlatepackAddress> for xDalekPublicKey {
+	type Error = Error;
```
yeastplume, line 102: Bonus point here for telling me why this conversion doesn't seem to work.

yeastplume:
Top comment updated, ready for review. Next PRs will be:

* Add API functions
* Add Slatepack flow to the command line as per the RFC

And then any tweaks to the format itself. Note we're going to be limited to deriving a single address per wallet for 4.0.0, and can figure out the details of multiple addresses when 4.0.0 is behind us.

libwallet/src/slatepack/address.rs
yeastplume: It does work, the ed25519 key just needs to be SHA512ed before using it as a decrypting key for x25519

PR #422, opened 2020-05-28
Fix Tor send

quentinlesceller:
Fixes #418.

PR #424, opened 2020-06-01
[3.1.x] Fix Tor send

quentinlesceller:
Equivalent of #422 for current/v3.1.x + version bump to 3.1.2.

Do Not Merge because it would be much better to release grin-wallet v3.1.2 with grin v3.1.1 which is still no published on crates.io.


quentinlesceller:
will merge once CI pass then release v3.1.2

PR #433, opened 2020-06-10
Add unpack command

yeastplume:
* Adds a `grin-wallet unpack` command, which unpacks a slatepack message, decrypts (if it's intended for this wallet), and displays the slate in JSON (if possible)
* Cleans up some of the command-line help text to reflect Slatepack changes


yeastplume:
> Looking good testing it now. What is the rationale for having an additional decrypt arg instead of trying to decrypt it by default?

Just that the logic flow was set up in most places to assume the decode function also handled decryption, was much easier to provide a flag to that function to allow it to handle the case where you just want the contents without decrypting than changing the flow everywhere.

yeastplume:
> Tested and everything looks fine.

Great, thanks for testing!

PR #434, opened 2020-06-10
[DNM] hardcode NRD kernel feature for testing

antiochp:
Hardcode NRD kernel features in wallet tx building code.
Every tx initiated with this wallet will use an NRD kernel with relative_height=60.

This NRD kernel variant is valid at HF3 if the feature flag is enable. This is a valid kernel feature variant is at block height 9 in usertesting chain type.

Note: This creates NRD kernels but does not allow for NRD kernels to be deterministic or reusable, each one has a random excess.

TODO - Hardcode the key and offset logic for deterministic reproduction of "known" excess values. We can do this through the use of a deterministic "random" excess during the key splitting, and offset as necessary based on the selected private keys (see the "two half kernels" test in node).

### DIFF
--- a/libwallet/src/slate.rs
+++ b/libwallet/src/slate.rs
@@ -372,29 +372,34 @@ impl Slate {
 	// 3: NRD (with associated relative_height)
 	// Any other value is invalid.
 	fn kernel_features(&self) -> Result<KernelFeatures, Error> {
-		match self.kernel_features {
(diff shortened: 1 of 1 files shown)


DavidBurkett:
@antiochp Are we good to close this now?

phyro:
@antiochp what's the state of this one, is it still relevant?

antiochp:
Closing. This was never intended to be merged. We just needed a way to temporarily test things.

PR #436, opened 2020-06-11
Slatepack Address HRP

yeastplume:
Calling it.

* `grin1` on mainnet
* `tgrin1` for floonet or any other testing chain type

(Last change before beta.2, and last intended functional change for 4.0.0 overall)

### DIFF
--- a/libwallet/src/slatepack/address.rs
+++ b/libwallet/src/slatepack/address.rs
@@ -241,6 +241,7 @@ impl Readable for SlatepackAddress {
 #[test]
 fn slatepack_address() -> Result<(), Error> {
 	use rand::{thread_rng, Rng};
+	global::set_local_chain_type(global::ChainTypes::AutomatedTesting);
 	let sec_key_bytes: [u8; 32] = thread_rng().gen();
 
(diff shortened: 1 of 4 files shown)


libwallet/src/slatepack/address.rs
```diff
@@ -43,8 +44,12 @@ pub struct SlatepackAddress {
 impl SlatepackAddress {
 	/// new with default hrp
 	pub fn new(pub_key: &edDalekPublicKey) -> Self {
+		let hrp = match global::get_chain_type() {
+			global::ChainTypes::Mainnet => "grin",
+			_ => "tgrin",
+		};
```
quentinlesceller, line 50: Does it make sense to have something else for AutomatedTesting/UserTesting (Regnet?) ?
Something like:
```suggestion
		let hrp = match global::get_chain_type() {
			global::ChainTypes::Mainnet => "grin",
			global::ChainTypes::Floonet => "tgrin",
			// UserTesting or AutomatedTesting
			_ => "rgrin",
		};
```

yeastplume: Possible, but my logic was that if someone's doing user testing or automated testing, it's in a local environment and they know what they're testing anyhow. If there's a call for it we can add another type in later.

PR #437, opened 2020-06-12
Version Bump for v4.0.0-beta.2

quentinlesceller:
As titled


quentinlesceller:
Waiting for tests to pass then I will merge and tag. Seems like Windows is a bit annoying today.

PR #440, opened 2020-06-15
Fix: height locked feature value should be 2.

jdwldnqi837:
// Match on kernel feature variant:
// 0: plain
// 1: coinbase (invalid)
// 2: height locked (with associated lock_height)
// 3: NRD (with associated relative_height)
// Anything else is unknown.


yeastplume:
Yes, that's correct, thanks for catching.

Issue #480, opened 2020-07-16
Removal of "finalize" params in InitSendTxArgs


quentinlesceller:
I just noticed that the `finalize` parameter was deleted from the `InitSendTxArgs ` and that by default the `init_send_tx` function will NOT `finalize` the transaction when sent with `InitSendTxArgs` and simply retrieves the initial transaction and as such S1 as slate state.

I think that either:
1. `init_send_tx` when `InitSendTxArgs` is present should `finalize` by default and retrieve the S3 transaction.
2. `init_send_tx` when `InitSendTxArgs` is present should returns the S2 transaction.

https://github.com/mimblewimble/grin-wallet/blob/master/api/src/owner.rs#L689-L704

also the outputs are not locked.

PR #485, opened 2020-07-20
TxLock and finalize

quentinlesceller:
Past behavior of the `init_send_tx` method with InitSendTxArgs (only when the recipient wallet was reachable):
1. Sender sends transaction to recipient
2. Recipients receives it and send back
3. Sender lock outputs
4. Sender `finalise` iff InitSendTxArgs contains `finalize=true` and returns the final slate
5. Sender `post_tx`  iff InitSendTxArgs contains `post_tx=true`  and returns the final slate

Current behavior with the same condition:
1. Sender sends transaction to recipient
2. Recipients receives it and send back
3. A. If `post_tx=true` Sender lock outputs, finalize and post_tx iff `post_tx=true`  and returns the final slate
3. B. If `post_tx=false` Sender returns the first slate

This PR locks outputs, finalise and return the final transaction but do not post when `post_tx=false`.

Fixes https://github.com/mimblewimble/grin-wallet/issues/480.

Might not be the best solution though. Thoughts @yeastplume ?

### DIFF
--- a/api/src/owner.rs
+++ b/api/src/owner.rs
@@ -701,7 +701,9 @@ where
 								}
 							}
 						} else {
-							return Ok(slate);
+							self.tx_lock_outputs(keychain_mask, &s)?;
+							let ret_slate = self.finalize_tx(keychain_mask, &s)?;
+							return Ok(ret_slate);
 						}
 					}
 					Ok(None) => Ok(slate),


quentinlesceller:
Do you think that'll be more logical to do something else like:

- `init_send_tx` when `InitSendTxArgs` is present should finalize by default and retrieve the S3 transaction. (this PR)
- `init_send_tx` when `InitSendTxArgs` is present should returns the S2 transaction.
- `init_sed_tx` reintroduce the `finalize` arg in `InitSendTxArgs` and returns S2 if finalize=false and S3 if finalize=true

yeastplume:
The send args are in place to emulate the wallet `send` command as closely as possible, so the idea was that if they're present, the command attempt to send, retrieve the S3 result and finalize in one call. So current behavior in this PR is consistent with how it's been working to date

quentinlesceller:
Also this is worth cherry picking into v4.0.x branch. Missed it somehow

PR #490, opened 2020-07-22
Post Hardfork 3 Cleanup

yeastplume:
Objective is to carefully remove the considerable amount of code and logic in places that was needed to support communication with V3.0.0 wallets pre-hardfork. Thus far (will update as we go):

* Removed V3 Slate and Versioning up/downgrade code
* Removed/tested command line switches and command line functions

Will likely not be merging this into the 4.0.x maintenance branch (will leave well-enough alone there, as these removals will need to be tested over time)


yeastplume:
Think this all seems to be okay now. Would appreciate if someone else could give it a quick sanity check before merging!

Issue #499, opened 2020-08-02
Regression: build_coinbase returns invalid output features


antiochp:
When mining, the grin node calls `build_coinbase()` on grin-wallet to build the coinbase output and kernel.

Grin node expects the output features to be in the form `"Coinbase"` but this is now returning `1`.
Edit: Also `com` and `prf` vs `commit` and `proof` (did not notice this initially).

https://github.com/mimblewimble/grin-wallet/blame/7a003374ac0d0189df0041edffaf332f4ded2d26/api/src/foreign_rpc.rs#L103

```
				"output": {
					"com": "08fe198e525a5937d0c5d01fa354394d2679be6df5d42064a0f7550c332fce3d9d",
					"features": 1,
					"prf": "9d8488fc..."
				}
```

https://github.com/mimblewimble/grin-wallet/blame/2888882e3086cc60b68499a8d9733fe4258b3f7d/api/src/foreign_rpc.rs#L106

```
				"output": {
					"commit": "08fe198e525a5937d0c5d01fa354394d2679be6df5d42064a0f7550c332fce3d9d",
					"features": "Coinbase",
					"proof": "9d8488fc..."
				}
```

antiochp:
I suspect `OutputFeaturesV4` is simply serializing to the values `0` or `1`.

I'm not actually sure if this is a larger issue - I saw this on `build_coinbase()` but not sure if there are other api calls between wallet and node that will be impacted.

In terms of fixing this - I'm not actually sure which would be the preferred serialization.
I can look into the serialization code on node and see if we can easily get serde to do the "right thing" here and support these simple integer values.

antiochp:
For reference, this is the error we see on node side when calling the wallet api to build the coinbase - 

```
20200802 15:25:32.048 ERROR grin_servers::mining::mine_block - 
Couldn't deserialize CbData: invalid type: integer `1`, expected string or map
```

PR #503, opened 2020-08-07
add "compat" output features

antiochp:
Related - #499.
Related - https://github.com/mimblewimble/grin/pull/3416.

V4 cleanup introduced a breaking change to the `CbData` response.

Old - 

```
"output": {
    "commit": "08fe198e525a5937d0c5d01fa354394d2679be6df5d42064a0f7550c332fce3d9d",
    "features": "Coinbase",
    "proof": "9d8488fc..."
}
```

New - 

```
"output": {
    "com": "08fe198e525a5937d0c5d01fa354394d2679be6df5d42064a0f7550c332fce3d9d",
    "features": 1,
    "prf": "9d8488fc..."
}
```

This PR ensures we now produce the expected json when node requests `CbData` via `build_coinbase`.

We actually no longer need `TransactionV4` and the various related data structures.
This PR cleans this up, allowing the use of a simplified `CoinbaseV4` with nested `CbOutputV4` and `CbKernelV4`.


antiochp:
I think we can actually clean a lot of this code up.
99% cure we no longer use `TransactionV4` at all and the only place we need "compat" features is in the `build_coinbase` api call from node.

I'll see if I can get something working and put a PR up.

PR #505, opened 2020-08-07
build_coinbase api fix and cleanup

antiochp:
Resolves: #499.
Related: https://github.com/mimblewimble/grin/pull/3416.

V4 cleanup introduced a breaking change to the `CbData` response.

Old - 

```
"output": {
    "commit": "08fe198e525a5937d0c5d01fa354394d2679be6df5d42064a0f7550c332fce3d9d",
    "features": "Coinbase",
    "proof": "9d8488fc..."
}
```

New - 

```
"output": {
    "com": "08fe198e525a5937d0c5d01fa354394d2679be6df5d42064a0f7550c332fce3d9d",
    "features": 1,
    "prf": "9d8488fc..."
}
```

This PR ensures we now produce the expected json when node requests `CbData` via `build_coinbase`.

We actually no longer need `TransactionV4` and the various related data structures.
This PR cleans this up, allowing the use of a simplified `CoinbaseV4` with nested `CbOutputV4` and `CbKernelV4`.

----

I was initially going to add an alternative "compat" ser/deser for output features along the same lines as the exsiting "compat" kernel features. But we do not actually need this as `CbData` and the version specific `CoinbaseV4` is the only place this would be required.


yeastplume:
All looks good, thanks for cleaning that up. Test failures look like CI's problem, so just triggering them again

PR #507, opened 2020-08-12
[WIP][DNM] Initial investigations into Late Locking

yeastplume:
Purpose of this PR is just to explore what we've been calling 'Lock-free transactions` but I think I'm going to just call 'Late Locking' for now, first outlined here https://forum.grin.mw/t/mostly-lock-free-transactions/7313. Will likely require an RFC if we aim to put this in production.

Preliminary work for this was done pre-HF4, so this is just an exploration of how easily it will be to implement and see what issues might come up. From what I can gather now, deciding how to handle the fee is the largest unknown.

WIP, and doesn't yet work, hoping to bring this to a point where a late-lock workflow is implemented in tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,7 +55,7 @@ checksum = "cfd7e7ae3f9a1fb5c03b389fc6bb9a51400d0c13053f0dca698c832bfd893a0d"
 dependencies = [
  "block-cipher-trait",
  "byteorder",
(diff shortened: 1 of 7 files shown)


Paouky:
Late Locking is definitely a better name to describe it clearly. Nice.

antiochp:
Closing. 
Superseded by #530.

PR #512, opened 2020-08-24
handle transaction inputs correctly

antiochp:
Related https://github.com/mimblewimble/grin/pull/3419

The node PR above defaults transactions to protocol version v3 which use "commit only" inputs, omitting the input features.

This PR updates how we create default empty transactions to ensure we continue to use "features and commit" inputs correctly.
Grin wallet serializes/deserializes transactions to the db in v1 format and we need to ensure we handle inputs correctly.


antiochp:
https://github.com/mimblewimble/grin/pull/3419 is now merged.
Updated Cargo.lock to reference latest sha on `master` for the various grin crates.
I will merge this PR once the tests pass.

PR #513, opened 2020-09-01
get_unspent api changed to take a commitment directly

antiochp:
Related: https://github.com/mimblewimble/grin/pull/3423

Fix call to `chain.get_unspent()`


antiochp:
Going to merge this to keep wallet in sync with grin on master.

Issue #516, opened 2020-09-09
"grin-wallet info" fails with "missing block height"


antiochp:
This was using grin-wallet built on `master` against a grin node also built against `master`.
The wallet itself is old and has not been touched for several months.

```
20200909 14:32:41.185 ERROR grin_wallet_impls::node_clients::http - 
Unexpected error in returned output (missing block height): 
Commitment(0985145c00fc6b0adef5d703c8028884493d73b3b2285b03c21d70fb0182d939d2). 
OutputPrintable { 
    output_type: Coinbase, 
    commit: Commitment(0985145c00fc6b0adef5d703c8028884493d73b3b2285b03c21d70fb0182d939d2), 
    spent: true, 
    proof: Some("3f14a385..."), 
    proof_hash: "fa16aa8e288a71df77c17e35ddea7d43ddb1a0cc36088316df7ef7ae43546f8e", 
    block_height: None, 
    merkle_proof: None, 
    mmr_index: 0 
}
```

Note that `spent: true`.

This appears to happen when parsing the response from the api call `get_unspent_outputs` which suggests we should only be expecting _unspent_ outputs when parsing it.

antiochp:
Actually this may be a local branch on my node with some changes to the MMR data, so I think I just realized why this is happening.

antiochp:
Closing. User error.

PR #517, opened 2020-09-11
we expect unspent outputs but play it safe

antiochp:
Related: #516 

We call the `get_unspent_outputs` api and expect to receive a collection of unspent outputs.
This is a reasonable assumption but we are brittle on the receiving/parsing side as each output in the response _must_ contain a block height. If we receive a _spent_ output for any reason the wallet is rendered inoperable as we can no longer refresh outputs.

This PR adds a (hopefully redundant) level of robustness by explicitly filtering out spent outputs from the list returned by the api.

See #516 for some background where the node api was inadvertently changed to return spent and unspent outputs, causing the wallet to fail.

### DIFF
--- a/impls/src/node_clients/http.rs
+++ b/impls/src/node_clients/http.rs
@@ -321,7 +321,8 @@ impl NodeClient for HTTPNodeClient {
 
 		let params = json!([start_index, end_index, max_outputs, Some(true)]);
 		let res = self.send_json_request::<OutputListing>("get_unspent_outputs", &params)?;
-		for out in res.outputs {
+		// We asked for unspent outputs via the api but defensively filter out spent outputs just in case.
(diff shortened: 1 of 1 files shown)


quentinlesceller:
Any idea why we are receiving a spent output? Shouldn't we filter them server side instead (or also)?

antiochp:
> Any idea why we are receiving a spent output? Shouldn't we filter them server side instead (or also)?

This was actually against a node running with code from https://github.com/mimblewimble/grin/pull/3428
So this was an unintentional break of the api contract (returned spent outputs when it should not).

But - I think we should filter these on both sides. The wallet should make as few assumptions as possible regarding data returned by the node.
So we should filter client side also.

quentinlesceller (approved):
👍 then

DavidBurkett (approved):

Issue #509, opened 2020-08-14
Owner API "can't create Tokio runtime"


JKKGBE:
**Describe the bug**
After authenticating with wallet's Owner API, when calling
```
node_height
init_send_tx (estimate)
init_send_tx
retrieve_txs
```
for some time in a loop, in an interval, where `init_send_tx` is to an offline wallet (for the time being I don't know if this happens when the recipient's wallet is online) using Slatepack address (again, don't know if address's type plays a role), the below happens:
```
$ ./grin-wallet --floonet owner_api
20200813 19:07:26.094 WARN grin_wallet_controller::controller - Starting HTTP Owner API server at 127.0.0.1:3420.
20200813 19:07:26.094 WARN grin_wallet_controller::controller - HTTP Owner listener started.
20200813 20:17:18.172 ERROR grin_wallet_impls::node_clients::http - Error calling get_tip: Internal error: can't create Tokio runtime
20200813 20:17:33.202 ERROR grin_wallet_impls::node_clients::http - Error calling get_tip: Internal error: can't create Tokio runtime
20200813 20:17:48.228 ERROR grin_wallet_impls::node_clients::http - Error calling get_tip: Internal error: can't create Tokio runtime
20200813 20:18:03.276 ERROR grin_wallet_impls::node_clients::http - Error calling get_tip: Internal error: can't create Tokio runtime
20200813 20:18:18.308 ERROR grin_wallet_impls::node_clients::http - Error calling get_tip: Internal error: can't create Tokio runtime
20200813 20:18:33.355 ERROR grin_wallet_impls::node_clients::http - Error calling get_tip: Internal error: can't create Tokio runtime
```
and calling `node_height` then gives `Ok.updated_from_node` as `false`, when before it was `true` every time. When trying to authenticate with the wallet's Owner API again, calling `init_secure_api` goes through, but `open_wallet` gives
```
{
  "code":-32099,
  "message":"Lifecycle: Error opening wallet: Wallet store error: LMDB error: Too many open files , Data Dir: /[...]/.grin/floo/wallet_data"
}
```

Restarting the wallet's Owner API fixes the problem until it happens again after ~1h 10m. The interval in the loop is 15s.

**Expected behavior**
No error.

**Desktop (please complete the following information):**
Operating System: Arch Linux
Kernel: Linux 5.7.12-arch1-1
Architecture: x86-64

Wallet: 4.0.0

**Additional context**
[...]

PR #520, opened 2020-09-30
[DNM] [5.0.x] Replace custom hyper client with reqwest

jaspervdm:
Right now our HTTP client that interfaces with the node and other wallets is using some pretty low level hyper code. It uses separate dependencies for tls and timeouts. On top of that, it uses a Tokio runtime that is spun up for every request separately, which is inefficient and tends to fail at some point (see https://github.com/mimblewimble/grin-wallet/issues/509).

This PR replaces the hyper code by the [reqwest](https://github.com/seanmonstar/reqwest) library, which is well reviewed and maintained and using it simplifies our code.

Instead of creating a tokio runtime for each request, use a global runtime. Unfortunately, until reqwest updates to tokio 0.3, the runtime will have to be behind a `Mutex`.


quentinlesceller:
This seems a great improvement 👍 .
> Could use some help with testing the tor socks proxy, if anyone feels like it.

Sure ping me anytime!

jaspervdm:
For some reason the tests are failing on Windows, I will try to debug it on my machine after the weekend as I don't have access to it before then.

quentinlesceller:
Looks like a random error. I triggered another run.

b-wg:
This seems good, any other issues with this PR? Can we merge it now?

phyro:
@quentinlesceller can we close this one?

quentinlesceller:
Yes indeed we can close this one.

PR #522, opened 2020-10-07
fix our lmdb iter usage to work with changes in grin

antiochp:
Our low level lmdb iter usage changed in https://github.com/mimblewimble/grin/pull/3450

This PR reworks the wallet usage of `db.iter()` to reflect these changes.

This is just to get everything compiling. We should revisit this as the interface is a little awkward right now.


antiochp:
The full walllet test suite ran locally for me. 
I'm guessing our test coverage of the various `iter()` calls in the db is relatively good here.

quentinlesceller (approved):
Looking good. Merging.

PR #526, opened 2020-10-27
fixfees on grin-wallet

tromp:
Complements the fixfees PR at https://github.com/mimblewimble/grin/pull/3481
Still needs some doc tests fixed, which requires some assistance from @yeastplume


util/Cargo.toml
```diff
@@ -35,20 +35,20 @@ sha3 = "0.8"
...
+# grin_store = { git = "https://github.com/mimblewimble/grin", branch = "master" }
 
 # For local testing
-# grin_core = { path = "../../grin/core"}
-# grin_keychain = { path = "../../grin/keychain"}
-# grin_chain = { path = "../../grin/chain"}
-# grin_util = { path = "../../grin/util"}
-# grin_api = { path = "../../grin/api"}
-# grin_store = { path = "../../grin/store"}
+grin_core = { git = "https://github.com/tromp/grin", branch = "fixfees" }
```
jaspervdm, line 46: We need to revert this file once the fixfees PR on grin is merged. Same with the Cargo.lock file

jaspervdm (commented):
This PR increases the fee of a tx, will this cause compatibility issues with older wallets? If an older wallet initiates a tx with a newer wallet, will the newer wallet deny it because the fee is too low?

Other than this issue (and the Cargo files) I think the PR looks good.

util/Cargo.toml
tromp: what you point out is not just a wallet issue. nodes running new code will not relay txs paying old fees.

jaspervdm (approved):
Upon further investigation there is no compatibility issue between upgraded and non-upgraded wallets, since the receiving side does not actually check the fee.

👍 from me once we fix the Cargo files after merging https://github.com/mimblewimble/grin/pull/3481.

api/src/owner_rpc.rs
```diff
@@ -256,7 +256,7 @@ pub trait OwnerRpc {
 			  "confirmation_ts": "2019-01-15T16:01:26Z",
 			  "confirmed": true,
 			  "creation_ts": "2019-01-15T16:01:26Z",
-			  "fee": null,
+			  "fee_fields": null,
```
quentinlesceller, line 259: Just so we know this is a breaking change with the current API.

quentinlesceller (commented):
Question about the API change below.

config/src/types.rs
```diff
@@ -75,6 +78,7 @@ impl Default for WalletConfig {
 			tls_certificate_key: None,
 			dark_background_color_scheme: Some(true),
 			keybase_notify_ttl: Some(1440),
+			accept_fee_base: None,
```
quentinlesceller, line 81: Shouldn't it be?
```suggestion
			accept_fee_base: default_accept_fee_base(),
```

controller/src/display.rs
```diff
@@ -412,17 +413,17 @@ pub fn estimate(
 		bMG->"Will be locked",
 	]);
 
-	for (strategy, total, fee) in strategies {
+	for (strategy, total, fee_fields) in strategies {
 		if dark_background_color_scheme {
 			table.add_row(row![
 				bFC->strategy,
-				FR->amount_to_hr_string(fee, false),
+				FR->amount_to_hr_string(fee_fields.fee(0), false),
```
quentinlesceller, line 420: Is it safe to use height 0 here?

libwallet/src/api_impl/owner.rs
```diff
@@ -402,8 +402,8 @@ where
 	let amount = if tx.amount_credited >= tx.amount_debited {
 		tx.amount_credited - tx.amount_debited
 	} else {
-		let fee = match tx.fee {
-			Some(f) => f,
+		let fee = match tx.fee_fields {
+			Some(f) => f.fee(0),
```
quentinlesceller, line 406: Same as above and actually it's kinda weird to use height 0 to have the value no? Wouldn't introduce a new function would be cleaner?

j01tz (commented):
:+1: once API conflict above is resolved. Will still require heavy testing in beta.

quentinlesceller (commented):
Question about the display of tx_fees below. Other than that I didn't catch anything.

config/src/types.rs
tromp: if accept_fee_base is not configured then it falls back to the global default. so these would seem to be equivalent.

api/src/owner_rpc.rs
jaspervdm: This also breaks compatibility with older wallets that already have tx log entries stored in their wallet db. This is not intended, will fix.

controller/src/display.rs
tromp: I guess hardfork4 height is better here.

j01tz (approved):

PR #527, opened 2020-10-28
refactor wallet init cli

gringineer:
When `grin-wallet init` is run `parse_init_args` throws an error if `grin-wallet.toml` already exists. 

This doesn't have to be the case.

`grin-wallet init` can still succeed if you have a`grin-wallet.toml` file without a `/wallet_data` directory.

You might run in to this if you edit/copy/paste your `grin-wallet.toml` from mainnet to floonet.

### DIFF
--- a/src/cmd/wallet_args.rs
+++ b/src/cmd/wallet_args.rs
@@ -22,7 +22,7 @@ use crate::util::{Mutex, ZeroingString};
 use clap::ArgMatches;
 use failure::Fail;
 use grin_wallet_api::Owner;
-use grin_wallet_config::{config_file_exists, TorConfig, WalletConfig};
+use grin_wallet_config::{TorConfig, WalletConfig};
 use grin_wallet_controller::command;
 use grin_wallet_controller::{Error, ErrorKind};
 use grin_wallet_impls::{DefaultLCProvider, DefaultWalletImpl};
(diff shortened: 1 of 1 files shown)


antiochp:
Hey @gringineer thanks for this. I'm in the process of getting back up to speed on wallet config, so might take a bit for me to fully understand what old/new behavior is here.

> You might run in to this if you edit/copy/paste your grin-wallet.toml from mainnet to floonet.

If you are setting a `testnet` (aka `floonet`) wallet up and you copy/paste a `mainnet` config file - what _should_ happen?

gringineer:
Hello @antiochp

**Ideally**
`grin init` won't fail if a `grin-wallet.toml` exists
`grin init` will fail if `./wallet_data/` exists

antiochp:
> `grin init` won't fail if a `grin-wallet.toml` exists

If we have a `grin-wallet.toml` that is `mainnet` specific (ports etc.) I have some concerns that we would _not_ want to silently succeed with `grin-wallet init` if we were actually init'ing a `testnet` wallet. This feels like it might be an accident waiting to happen.

What would happen with this PR in this case?

gringineer:
Why wouldn't the command line argument override `grin-wallet.toml`?

```
if command_line_variables:
    return command_line_variables
else if config_file_variables:
    return config_file_variables
else:
    return hard_code_value
```

Does `grin-wallet` not follow the above logic?

gringineer:
Gonna close as I have figured out a way around the block. 

:+1:

Issue #529, opened 2020-11-05
Add ability to select outputs for send


jseibel-circle:
**Is your feature request related to a problem? Please describe.**
When the amount of unspent outputs (UTXOs) in the wallet gets large, the wallet starts to time out and run slowly. If we had the ability to explicitly select outputs, we could keep the total number of UTXOs more strategic and have better logic on how to fund sends to reduce fees.

**Describe the solution you'd like**
I would like a new flag for the `send` command that allowed a list of output commitments. Commitment hash is long but it allows you to be super explicit. It can either be a single flag where you pass string comma-separated or pass the same flag numerous times. Should gracefully handle this even if very high number of selections (over 100). Also update the API.

**Describe alternatives you've considered**
We are trying to get the existing `smallest` selection strategy to work but it is cumbersome. We could also craft our own transaction manually but there is no easy way to sign a raw tx with the private key, we would have to reverse engineer the wallet code.

antiochp:
Totally agree. This effectively exposes the ability to use "custom" selection strategies with external logic.

antiochp:
There is some interesting interaction with "late locking" (see #530) here that I don't think we have considered.

Given how we are thinking about "late locking" we would not actually be selecting the specific outputs during initial transaction creation. We would specify the tx fee which in turn would implicitly constrain the max number of outputs spent by the transaction.
And then during _finalize_ we would choose the specific outputs to spend, given the constraints.

i.e. Define a large enough fee on the initial transaction and we can choose 500 outputs to spend during `finalize`, assuming they cover the amount being sent. We then create the necessary change output to balance the transaction.

Need to think about this a bit more.

PR #530, opened 2020-11-11
Late locking (experimental)

jaspervdm:
Adding experimental late locking feature to the wallet. If this option is set during tx creation, no UTXOs will be locked on the wallet. Instead, they will only be locked as late as possible, namely when the transaction is finalized.

This feature is considered experimental. If during finalization we realize that the initial fee is not sufficient to pay for the transaction anymore, we simply return an error. This scenario will probably be rare for regular users, but wallets that have many transactions outstanding simultaneously might encounter this. The solution to this would be to add another kernel that makes up the fee difference. This will be addressed in a future PR.

Thanks to @yeastplume for helping with this new feature.

TODOs:
- ~~Remove the `3*fee` and output a nice warning when the final fee does not match the initial fee~~
- ~~Add CLI flag to enable late locking~~


Paouky:
Great!
Do you see it as becoming the default option once the 2 items on the TODO list are crossed?

antiochp:
> Great!
> Do you think it should be set as the default option once these 2 items on the TODO list are crossed?

Eventually yes, but there is a fair amount of testing and simply trying this out in various scenarios and under various conditions before we would consider making it the default.

libwallet/src/api_impl/owner.rs
```diff
@@ -667,6 +679,19 @@ where
 	)?;
 
 	let keychain = w.keychain(keychain_mask)?;
+
+	// Add our contribution to the offset
+	if context_res.is_ok() {
+		// Self sending: don't correct for in- and outputs here
+		//  as we will do it during finalization.
```
antiochp, line 686: Nit: Extra whitespace.

libwallet/src/api_impl/owner.rs
```diff
@@ -667,6 +679,19 @@ where
 	)?;
 
 	let keychain = w.keychain(keychain_mask)?;
+
+	// Add our contribution to the offset
+	if context_res.is_ok() {
+		// Self sending: don't correct for in- and outputs here
```
antiochp, line 685: Typo? Maybe just "inputs and outputs".

libwallet/src/api_impl/owner.rs
```diff
@@ -754,16 +778,51 @@ where
...
 	let parent_key_id = w.parent_key_id();
 
-	// since we're now actually inserting our inputs, pick an offset and adjust
-	// our contribution to the excess by offset amount
-	// TODO: Post HF3, this should allow for inputs to be picked at this stage
-	// as opposed to locking them prior to this stage, as the excess to this point
-	// will just be the change output
+	if let Some(args) = context.late_lock_args.take() {
+		// Transaction was late locked, select inputs+change now
+		//  and insert into original context
```
antiochp, line 787: Nit: extra whitespace.

controller/tests/late_lock.rs
```diff
@@ -0,0 +1,163 @@
...
+			late_lock: Some(true),
+			..Default::default()
+		};
+		let slate_i = sender_api.init_send_tx(m, args)?;
+		println!("S1 SLATE: {}", slate_i);
+		slate = client1.send_tx_slate_direct("wallet2", &slate_i)?;
+		println!("S2 SLATE: {}", slate);
+
+		// Note we're not doing this here
+		// sender_api.tx_lock_outputs(m, &slate)?;
```
antiochp, line 118: Maybe remove the commented code and replace with a comment to explain what is happening?
Or maybe this should just be removed entirely?

antiochp (commented):
Still reviewing. 
Commented on a couple of very tiny cleanups in comments etc.

Looks :+1: so far.

libwallet/src/api_impl/owner.rs
jaspervdm: This is on purpose, because my IDE detects multi line sentences without an extra space in the beginning as a separate comment. 
This is how it highlights a TODO comment with and without an extra space:
IDE screenshot
But I'll change it anyway, since I don't think we do that anywhere else in the codebase.

jaspervdm:
Reolved the TODO items. Currently the finalization step errors if the intial fee is different from the final fee. Technically we could handle the case where the initial fee is _higher_ than the final fee with relative ease. It means we slightly overpay for the tx but it would be better than the tx failing to finalize.

antiochp:
> Currently the finalization step errors if the intial fee is different from the final fee.

This is probably fine to go with for now? I'm assuming this would be a minor change that affects only a single wallet (the one finalizing the tx) so rolling out an improvement for this would be a minor task.

antiochp (approved):
:+1:
This is cool.
Tested locally generating a bunch of "competing" slatepacks with different sends and able to choose any one of them to actually receive and finalize.

antiochp:
"Coin control" gets interesting with "late locking" (see #529).

We need to specify a large enough fee up front, but we do not need to select explicit outputs to spend until the `finalize` step.

libwallet/src/api_impl/owner.rs
```diff
@@ -754,16 +778,51 @@ where
...
-	// as opposed to locking them prior to this stage, as the excess to this point
-	// will just be the change output
+	if let Some(args) = context.late_lock_args.take() {
+		// Transaction was late locked, select inputs+change now
+		// and insert into original context
+
+		let current_height = w.w2n_client().get_chain_tip()?.0;
+		let mut temp_sl =
+			tx::new_tx_slate(&mut *w, context.amount, false, 2, false, args.ttl_blocks)?;
+		let temp_context = selection::build_send_tx(
```
antiochp, line 792: I think there may be an issue here with `build_send_tx()` when late locking.
We take the amount from the slate, not from the local private ctx.

antiochp: We restore the original fee and amount on the slate based on out local ctx in `repopulate_tx()` but this happens _after_ we build the tx (based on the slate) during late locking finalization.

antiochp: Presumably some subset of data is "set once", transmitted via the slate, but should always be read from the local ctx by either party once set.
Ideally this data is not even included in the slate in subsequent communication.

antiochp: Quick summary of discussion - 

When we call `build_send_tx()` we pass in the _temporary_ slate `temp_sl`, not the slate passed from the recipient.
This temporary slate contains `amount` from our local `context`.

We then update `context` with input and output ids from the _temporary_ context `temp_context` created by `build_send_tx()`.
The updated private `context` is then persisted locally.

PR #531, opened 2020-11-19
default to smallest strategy for paying invoices

antiochp:
We use the `smallest` coin selection strategy by default when sending funds.
But we were defaulting to `all` when paying an invoice.

This PR standardizes both payments to use `smallest` consistently.


jaspervdm (approved):
LGTM :+1+:

This would also put it in line with behaviour on the owner API, which already defaults to smallest.

PR #532, opened 2020-11-20
Deprecate HTTP(S)

jaspervdm:
Implementation of the [Deprecate HTTP(S) Transactions RFC](https://github.com/mimblewimble/grin-rfcs/pull/54).

### DIFF
--- a/impls/src/adapters/mod.rs
+++ b/impls/src/adapters/mod.rs
@@ -22,9 +22,8 @@ pub use self::http::{HttpSlateSender, SchemeNotHttp};
 pub use self::keybase::{KeybaseAllChannels, KeybaseChannel};
 pub use self::slatepack::PathToSlatepack;
 
-use crate::config::{TorConfig, WalletConfig};
-use crate::libwallet::{Error, ErrorKind, Slate};
-use crate::tor::config::complete_tor_address;
+use crate::config::WalletConfig;
+use crate::libwallet::{Error, Slate};
 use crate::util::ZeroingString;
 
 /// Sends transactions to a corresponding SlateReceiver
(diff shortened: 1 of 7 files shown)


impls/src/adapters/mod.rs
```diff
@@ -59,64 +58,3 @@ pub trait SlateGetter {
...
-				return Err(
-					ErrorKind::WalletComms("Tor Configuration required".to_string()).into(),
-				);
-			}
-			Some(tc) => Box::new(
-				HttpSlateSender::with_socks_proxy(&dest, &tc.socks_proxy_addr, &tc.send_config_dir)
-					.map_err(|_| invalid())?,
-			),
-		},
-		"keybase" => Box::new(KeybaseChannel::new(dest)?),
```
j01tz, line 101: With this change, is there any reason to import or keep around any of the keybase logic in adapters?

j01tz (approved):
Looks good overall, thanks for taking care of this @jaspervdm.

I ran it on a test instance with Tor not installed and got following output

20201124 09:45:40.939 WARN grin_wallet_controller::controller - Starting TOR Hidden Service for API listener at address rglop53gkfop22lfka5z2juxsv4f6uciwseg7dslsay72otevsrsupyd, binding to 127.0.0.1:13415
20201124 09:45:40.944 WARN grin_wallet_controller::controller - Unable to start TOR listener; Check that TOR executable is installed and on your path
20201124 09:45:40.944 WARN grin_wallet_controller::controller - Tor Error: Tor Process Error: Process("TOR executable (`tor`) not found. Please ensure TOR is installed and on the path: Os { code: 2, kind: NotFound, message: \"No such file or directory\" }")
20201124 09:45:40.944 WARN grin_wallet_controller::controller - Listener will be available via HTTP only
20201124 09:45:40.945 WARN grin_wallet_controller::controller - Starting HTTP Foreign listener API server at 127.0.0.1:13415.
20201124 09:45:40.945 WARN grin_wallet_controller::controller - HTTP Foreign listener started.
I think it is doing what we expect but a couple of usability things that may be worth noting for a future PR:

- do we want the first log line above to return a tor address or a slatepack address? this could add confusion to return a raw tor address (or am I missing something in how the API is done?)
- even though http is no longer supported, the warning logs can confuse a user into thinking http is still working

antiochp (approved):
Looks good to me. :+1:

impls/src/adapters/mod.rs
jaspervdm: Afaik we still support Keybase as a method. This particular block of code was just unused so I got rid of it entirely.

jaspervdm:
@j01tz 
(i edited your comment by accident instead of replying to it, my bad)

> even though http is no longer supported, the warning logs can confuse a user into thinking http is still working

Technically we still have a HTTP listener, it just runs locally. This listener is also used by the TOR hidden service. But I do agree we could improve on the text here in a future PR.

PR #535, opened 2020-11-26
Bump version to 5.0.0-beta.1

jaspervdm:
- `4.1.0-alpha.1` -> `5.0.0-beta.1`
- Update grin dependency to `5.0.0-beta.1`

PR #536, opened 2020-11-26
Bump master version to 5.1.0-alpha.1

jaspervdm:
Equivalent of https://github.com/mimblewimble/grin-wallet/pull/535 but for master branch.
- `4.1.0-alpha.1` -> `5.1.0-alpha.1`
- Update grin dependency to `master`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1170,8 +1170,8 @@ checksum = "9b919933a397b79c37e33b77bb2aa3dc8eb6e165ad809e58ff75bc7db2e34574"
 
 [[package]]
 name = "grin_api"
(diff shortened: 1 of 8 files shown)


antiochp:
If we are bumping version on `master` we should bump it to `5.1.0-alpha.1` to differentiate it from the `5.0.x` branch.
Probably not a big deal right now but it will be useful if we need to patch the `5.0.0` between now and HF4.

antiochp:
> Update grin dependency to 5.0.0-beta.2

We want to keep `master` pointing at `master` so we can keep these updated easily over time.

antiochp (approved):
:+1:

PR #538, opened 2020-11-27
fix yet another instance of fee fields at height 0

tromp:
The former aggregate_fee_fields(0) call would fail on huge fees.
The fixed one will always apply the fee mask.
It can still fail on a tx with multiple kernels whose fees add up to a huge one.
Merging this is not that urgent. Merge whenever convenient.

### DIFF
--- a/libwallet/src/api_impl/owner.rs
+++ b/libwallet/src/api_impl/owner.rs
@@ -907,7 +907,7 @@ where
 		Some(tx) => {
 			let mut slate = Slate::blank(2, false);
 			slate.tx = Some(tx.clone());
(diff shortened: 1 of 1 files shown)


antiochp:
> Merging this is not that urgent.

We still want to get this in for 5.0.0 right?

tromp:
> We still want to get this in for 5.0.0 right?

We already have 5.0.0
I think it's not worth rolling out a new 5.0.1 just for this.

DavidBurkett (approved):
@tromp Looks good to me. Do you have privs to merge this?

tromp:
no, i don't have merge priviledge.

antiochp (approved):
:+1:

PR #542, opened 2020-12-09
post late locked tx and assert amount on both wallets

antiochp:
This test was never actually posting the tx to the node.
The tx was never included in a block. Funds were never received by the receiving wallet.

* post late locked tx
* assert amount on both wallets
* cleanup wallet account setup


quentinlesceller (approved):
Looking good. Testing locally now.

quentinlesceller:
Tested. LGTM

PR #543, opened 2020-12-10
Use global Tokio runtime

jaspervdm:
Splitting the actual bugfix off from https://github.com/mimblewimble/grin-wallet/pull/520. Fixes https://github.com/mimblewimble/grin-wallet/issues/509.

Instead of creating a tokio runtime for each request, use a global runtime. Unfortunately, until reqwest updates to tokio 0.3, the runtime will have to be behind a `Mutex`.


antiochp (commented):
Makes sense to me at a high level.
Ideally somebody with more experience with Tokio can take a look as well.

I don't think the mutex is too much of an issue. I'm guessing most requests are one-by-one anyway and that there will unlikely be much contention (if any).

I'll plan to do some local testing with this today.

antiochp (approved):
:+1:

PR #544, opened 2020-12-14
replace api_secret with foreign_secret + naming of APIs files and variables

deevope:
- Fix of [bladedoyle issue](https://github.com/Grinnode-live/2020-grin-bug-bash-challenge/issues/15) 
- Changed name of `.api_secret` to `.node_foreign_api_secret` and `.owner_api` to `.wallet_owner_api_secret`
- Better naming of wallet `owner_api` and node `foreign_api` variables


quentinlesceller:
Can you rebase your PR? I think this should be nice to have,

phyro:
does this _require_ specific node version where it has this https://github.com/mimblewimble/grin/pull/3525 patch? meaning that upgrading the wallet without upgrading the node would not work? not sure if this is a problem though

deevope:
Sorry for the delay, I was quite busy lately.

> Can you rebase your PR? I think this should be nice to have,

Sure I will do.

> does this _require_ specific node version where it has this [mimblewimble/grin#3525](https://github.com/mimblewimble/grin/pull/3525) patch? meaning that upgrading the wallet without upgrading the node would not work? not sure if this is a problem though

Right now, yes. I need to make it backward compatible otherwise the wallet will create `.node_foreign_api_secret` and it couldn't be detected by default by the node.

Talking about this, I've 2 ways to make it backward compatible:
1. Rename existing file  `.api_secret ` to  `.node_foreign_api_secret ` for example
2. If `.api_secret ` exist, we take it as default in the config otherwise we create `.node_foreign_api_secret ` file as usual workflow in the wallet.

I think the solution 2 is the best for the users given that it doesn't make conflict with possible existing usage of this old api secret file on other programs.
I'll go with 2nd then?

phyro:
I agree with your points and find option 2 having better backwards compatibility 🤷‍♂️

phyro:
I suggest we move with the option 2 in order to avoid getting this PR stuck in time. It will likely get more reviews once we have one possible solution ready to be reviewed/merged :+1:

phyro:
I think this one still needs to be rebased because there's quite a few commits here that I would not expect. It could even be that our master is not yet up to date with some changes 🤔

deevope:
> I think this one still needs to be rebased because there's quite a few commits here that I would not expect. It could even be that our master is not yet up to date with some changes 🤔

Indeed, sorry didn't make attention to that. Done, I just need to make sure everything is working as expected tonight and it will be ready for a review 👍

deevope:
@quentinlesceller It should be good for merging right now 👍 Some things to note right know is if the `grin-wallet.toml` is deleted we chose the old foreign/owner file if they exist. The best would be to look at the `grin-server.toml` to know what is the current path for the foreign api secret but since this has been implement, I believe it's fine as it is right now. Also, I noticed if we delete the `grin-wallet.toml` and we don't redo the init command to recreate a wallet it will take the default config. Shouldn't we try to always have a `grin-wallet.toml`? If yes, could i include it in this pr  or i should make another pr related to this issue if it's one.

deevope:
@quentinlesceller ready for review 👍

deevope:
Well, you can test and merge it. Users will only encounters some issue if they delete the old API files, it will recreate one with the new name but it will not be in function of the old `grin-wallet.toml`. The commit #00f1428 mitigate this but the test is failing, i could fix it using the api_secret or disabling it in the test. 
But still want some opinion which solution to chose : 
- Soft compatibility, test and merge
or
- Hard compatibility, fix the test, test and merge

deevope:
Decided to swith to "Hard compatibility". There's still some issue persisting : 

- For example, if a user delete old `grin-server.toml` and `.foreign_api_secret`, launch directly the node and after the grin-wallet, the node it will create the new foreign api secret file name `.node_foreign_api_secret` on the node side and the wallet will create the old foreign api secret file name `.foreign_api_secret`. And it would lead the grin-wallet can't communicate with node until the user delete the old api_secret_file name and `grin-wallet.toml`. But I think this issue, is rare enough to not bother about this. 
- Wallet like niffler would need to support both api secret file name

Otherwise, it should be good. All my test were successful.

Issue #528, opened 2020-11-05
wallet --estimate-selection throws an error


antiochp:
```
./target/release/grin-wallet send 0.1 --estimate-selection

Estimation for sending 0.100000000:

+--------------------+-------------+----------------+
| Selection strategy | Fee         | Will be locked |
+====================+=============+================+
| smallest           | 0.007000000 | 0.407000000    |
+--------------------+-------------+----------------+
| all                | 0.005000000 | 5.000000000    |
+--------------------+-------------+----------------+

20201105 15:01:08.345 DEBUG grin_wallet_api::owner - Send (TOR): Destination is not SlatepackAddress Error { inner: 

SlatepackAddress error: invalid length }
20201105 15:01:08.345 DEBUG grin_wallet_api::owner - Send (HTTP): Cannot create HTTP Slate sender SchemeNotHttp
Wallet command failed: LibWallet Error: Wallet store error: DB Not Found Error: Slate id: [...]
```

antiochp:
The error suggests we do not exit early when using `--estimate-selection` and inadvertently try and do something subsequent with the slate.

PR #545, opened 2020-12-15
Fix error when estimating fees for selection strategies

jaspervdm:
Fixes https://github.com/mimblewimble/grin-wallet/issues/528.


antiochp (approved):
I'm good with this if we are comfortable with it not introducing any seemingly unrelated issues.
Both changes are gated by `estimate_selection_strategies` bool.

PR #546, opened 2020-12-15
[5.0.x] replace .api_secret to .foreign_api_secret

deevope:
Fix of bladedoyle issue Grinnode-live/2020-grin-bug-bash-challenge#15


jaspervdm:
Thanks! Could you PR this against the current/5.0.x branch so we can include it in the next beta and onwards?

deevope:
Done! 👌

quentinlesceller (approved):
Tested from scratch and works well.

PR #562, opened 2021-01-04
Downgrade CI VM to ubuntu-18.04

DavidBurkett:


antiochp (approved):
:+1:

antiochp:
Did we have builds failing due to this? Do we need to port this across to the `5.0.x` branch also?

DavidBurkett:
No builds were failing AFAIK. Without this though, the 5.0.0 releases won't be very portable.

PR #565, opened 2021-01-05
Allowing the creation of height locked Slates Fixes #564

### LINKED ISSUE #564: Unable to set the `kernel_features_args` in Slate object
**Describe the bug**
It is currently not possible to set the `kernel_features_args` of a Slate object because the field is of type `Option<KernelFeaturesArgs>` while `KernelFeaturesArgs` is only available in the private `slate` module. Furthermore the widely used `pub fn blank(num_participants: u8, is_invoice: bool) -> Slate` used to initialize a slate will just set this field to `None`
image

**To Reproduce**
```Rust
use grin_wallet_libwallet::{ Slate, slate::KernelFeaturesArgs };
let kfeat = KernelFeatures::HeightLocked{
    fee : fee_field,
    lock_height: timelock
};
let fee = 500000;
let lock_height = 500000;
let fee_field = FeeFields::new(0, fee).unwrap();
slate.kernel_features = kfeat.as_u8();
slate.kernel_features_args = KernelFeaturesArgs {
    lock_height: lock_height
}
```

**Expected behavior**
Kernel feature args should be settable somehow (either at slate initialization or with a setter)

**Screenshots**
image
image

 - Ubuntu
 - master branch

**Additional context**
This is a problem for wallets who want to implement height locked transactions.

jafalter:
Created new function `blank_with_kernel_features` which takes (compared to the regular `blank(num_participants: u8, is_invoice: bool))` the additional parameters `kernel_feat_param: u8,` and `lock_height_param: Option<u64>` such that one can create a slate with a height locked kernel.

This fixes the issue https://github.com/mimblewimble/grin-wallet/issues/564

I have adjusted `blank(num_participants: u8, is_invoice: bool)` to use this new function to avoid code duplication.

### DIFF
--- a/libwallet/src/slate.rs
+++ b/libwallet/src/slate.rs
@@ -250,8 +250,21 @@ impl Slate {
 		tx
 	}
 
-	/// Create a new slate
+	/// Create a new slate with a plain kernel
 	pub fn blank(num_participants: u8, is_invoice: bool) -> Slate {
+		Slate::blank_with_kernel_features(num_participants, is_invoice, 0, None)
+	}
+
+	/// Create a new slate with custom kernel_feature and lock height
+	/// If a lock_height is set kernel_features will have to be either HeightLocked (2) or NoRecentDuplicate (3)
+	/// otherwise the value passed as lock_height will be ignored and a plain kernel will be created
+	/// Invalid arguments for kernel_features (>3) will be set to 0 (Plain)
+	pub fn blank_with_kernel_features(
+		num_participants: u8,
+		is_invoice: bool,
+		kernel_feat_param: u8,
+		lock_height_param: Option<u64>,
+	) -> Slate {
 		let np = match num_participants {
 			0 => 2,
 			n => n,
@@ -259,6 +272,20 @@ impl Slate {
 		let state = match is_invoice {
 			true => SlateState::Invoice1,
 			false => SlateState::Standard1,
+		};
+		let kernel_features = if kernel_feat_param > 3 {
+			0
+		} else {
+			kernel_feat_param
+		};
+		let is_height_locked =
+			lock_height_param.is_some() && (kernel_features == 2 || kernel_features == 3);
+		let kernel_feat_args = if is_height_locked {
+			Some(KernelFeaturesArgs {
+				lock_height: lock_height_param.unwrap(),
+			})
+		} else {
+			None
 		};
 		Slate {
 			num_participants: np, // assume 2 if not present
@@ -268,7 +295,7 @@ impl Slate {
 			amount: 0,
 			fee_fields: FeeFields::zero(),
 			ttl_cutoff_height: 0,
-			kernel_features: 0,
+			kernel_features: kernel_features,
 			offset: BlindingFactor::zero(),
 			participant_data: vec![],
 			version_info: VersionCompatInfo {
@@ -276,9 +303,10 @@ impl Slate {
 				block_header_version: GRIN_BLOCK_HEADER_VERSION,
 			},
 			payment_proof: None,
-			kernel_features_args: None,
+			kernel_features_args: kernel_feat_args,
 		}
 	}
+
 	/// Removes any signature data that isn't mine, for compacting
 	/// slates for a return journey
 	pub fn remove_other_sigdata<K>(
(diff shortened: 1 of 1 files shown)


DavidBurkett (approved):
This change looks good to me. @antiochp @quentinlesceller, can one of you merge this?

libwallet/src/slate.rs
```diff
@@ -260,6 +273,20 @@ impl Slate {
 			true => SlateState::Invoice1,
 			false => SlateState::Standard1,
 		};
+		let kernel_features = if kernel_feat_param > 3 {
+			0
```
pkariz, line 277: my guess is that `kernel_feat_param` being > 3 means there's a bug, so is it possible that a panic would be better in this case?

phyro:
@antiochp @quentinlesceller do you guys think this one is good to merge?

libwallet/src/slate.rs
jafalter: agreed, fixed

libwallet/src/slate.rs
```diff
@@ -250,8 +250,21 @@ impl Slate {
...
-	/// Create a new slate
+	/// Create a new slate with a plain kernel
 	pub fn blank(num_participants: u8, is_invoice: bool) -> Slate {
+		Slate::blank_with_kernel_features(num_participants, is_invoice, 0, None)
+	}
+
+	/// Create a new slate with custom kernel_feature and lock height
+	/// If a lock_height is set kernel_features will have to be either HeightLocked (2) or NoRecentDuplicate (3)
+	/// otherwise the value passed as lock_height will be ignored and a plain kernel will be created
+	/// Invalid arguments for kernel_features (>3) will be set to 0 (Plain)
```
phyro, line 261: I think we should update this comment as well now that we made it panic

jafalter: very true, thanks! Corrected the comment now.

phyro:
alright, @antiochp @quentinlesceller could one of you two also give it a quick look now and merge if it looks good?

libwallet/src/slate.rs
```diff
@@ -250,8 +250,21 @@ impl Slate {
...
+	/// Create a new slate with custom kernel_feature and lock height
+	/// If a lock_height is set kernel_features will have to be either HeightLocked (2) or NoRecentDuplicate (3)
+	/// otherwise the value passed as lock_height will be ignored and a plain kernel will be created
+	/// Invalid arguments for kernel_features (>3) are not allowed and will make the function panic
+	pub fn blank_with_kernel_features(
+		num_participants: u8,
+		is_invoice: bool,
+		kernel_feat_param: u8,
+		lock_height_param: Option<u64>,
+	) -> Slate {
```
quentinlesceller, line 267: Would that makes sense to pass a kernel_features like that?
```suggestion
	pub fn blank_with_kernel_features(
		num_participants: u8,
		is_invoice: bool,
		kernel_features: KernelFeatures,
	) -> Slate {
```

That would avoid to have the whole parsing and possible panic in here.

quentinlesceller (commented):
Thank you for the PR @jafalter! I have a question regarding safety and possible simplification.

libwallet/src/slate.rs
phyro: I agree with this suggestion, doing it in this way also doesn't need code modification if/when we add new kernel features :+1:

phyro: @jafalter pinging just in case it was forgotten

jafalter: Thanks, I will do that. But I just need to doublecheck because I remember one of the types was not publicly visible I think that is why I did it like that.

jafalter: @quentinlesceller @phyro the Slate struct requires a `Option<KernelFeaturesArgs>` see:
image
`KernelFeaturesArgs` is not a publicly visible struct, therefore can't be passed as the parameter. Using `KernelFeatures` as you recommended would require bigger refactoring of the Slate struct.

phyro: @jafalter sorry for the late response. I see, in this case I'm ok if we go this way and we can possibly refactor this later :+1:
cc: @quentinlesceller

cekickafa:
is that forgotten?

wiesche89:
Replaced by #793, which also includes the review feedback and V4 roundtrip tests.

Issue #559, opened 2021-01-04
initiating scan via Owner API on v5 wallet is causing 'tokio-runtime-worker' panic


marekyggdrasil:
**Describe the bug**

During the [grinnode.live Winter 2020 Bug Bash Challenge](https://forum.grin.mw/t/grinnode-live-winter-2020-bug-bash-challenge-and-free-grin-btc-prizes-giveaway/8060) a tester @mojitoo discovered a problem with the wallet. He was able to cause v5 wallet to `'tokio-runtime-worker' panic` by initiating scan via Owner API. It was discovered while working on test case https://github.com/Grinnode-live/2020-grin-bug-bash-challenge/issues/37. His detailed info is provided in [gist](https://gist.github.com/mojitoo/884d628029f24060bfd26ed705023890) and I managed to reproduce the problem [on my own](https://gist.github.com/marekyggdrasil/c953f486cdeb9a7fba8de8542b783e9d)

**To Reproduce**

1. Prepare a v5 wallet and put it in Owner API listener mode.
2. Make an API call to run a scan method.
3. You should see `'tokio-runtime-worker' panic` in wallet v5 logs.

Bug will be reproduced if you follow steps from https://github.com/Grinnode-live/2020-grin-bug-bash-challenge/issues/37 test case.

**Expected behavior**

We would expect the `'tokio-runtime-worker' panic` to not occur.

**Screenshots**

N/A

**Desktop (please complete the following information):**

Tester environment
```
Darwin MacBook-Pro-de-Workstation.local 19.5.0 Darwin Kernel Version 19.5.0: Tue May 26 20:41:44 PDT 2020; root:xnu-6153.121.2~2/RELEASE_X86_64 x86_64
```
Reproduction environment
```
Linux 4.19.0-12-amd64 #1 SMP Debian 4.19.152-1 (2020-10-18) x86_64 unknown unknown GNU/Linux
```

**Additional context**

[Tester report](https://gist.github.com/mojitoo/884d628029f24060bfd26ed705023890) and my report on [reproducing the bug](https://gist.github.com/marekyggdrasil/c953f486cdeb9a7fba8de8542b783e9d)

jaspervdm:
Thanks for reporting! We fixed this issue in other places shortly before releasing 5.0 but apparently missed this one. I have a PR up now that fixes it.

PR #573, opened 2021-01-13
[5.0.x] Fix owner API scan

jaspervdm:
When introducing the fix in https://github.com/mimblewimble/grin-wallet/pull/552 we missed the scan api method, since it uses a `FuturesUnordered` instead of `Client::send_request`.

Fixes https://github.com/mimblewimble/grin-wallet/issues/559.


quentinlesceller (approved):
Tested and this fix common issues encoutered with the wallet owner api. LGTM and release v5.0.3.

PR #596, opened 2021-03-17
Clippy easy wins

trevyn:
To make these easy to review, these are nice obvious wins that make the code smaller and cleaner. :)

All of the removed `clone()`s are unnecessary, usually because that type implements `Copy`.

`cargo test --all` passes locally.


quentinlesceller (approved):
Looking good! Thank you @trevyn. Will merge once CI pass.

PR #602, opened 2021-04-05
remove height argument from fee related functions as per fixpastfees RFC

tromp:
name: fixpastfees
about: remove dependence of fee() and fee_shift() functions on height
title: 'fix past fees'

Companion PR for

https://github.com/mimblewimble/grin/pull/3629

both implementing RFC https://github.com/mimblewimble/grin-rfcs/blob/master/text/0021-fix-prior-fees.md
to extend the 40 bit fee restriction and use of fee shifts back beyond HF4 to all history.
Technically a Hard Fork, but if we have a many month reorg introducing a > 40 bit fee prior to HF4, we have bigger problems to deal with than a chain split.

This simplifies all fee related functions as they no longer need a height argument to distinguish preHF4 from postHF4 behaviour.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1180,8 +1180,8 @@ checksum = "9b919933a397b79c37e33b77bb2aa3dc8eb6e165ad809e58ff75bc7db2e34574"
 
 [[package]]
 name = "grin_api"
-version = "5.1.0-alpha.1"
-source = "git+https://github.com/mimblewimble/grin?branch=master#f6ec77a592de04fcc4a78127822932ec7e18525c"
+version = "5.2.0-alpha.1"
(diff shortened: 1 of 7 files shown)


phyro:
@tromp this one seems to have some failing tests

tromp:
how to add revieers again?

GeneFerneau:
Had to run a `cargo update` to bump the `master` version locally, maybe the CI needs it too?

GeneFerneau:
`git add Cargo.lock`?

GeneFerneau (approved):

DavidBurkett (approved):

quentinlesceller (approved):
LGTM 👍

PR #603, opened 2021-04-05
Fix Windows CI `cargo test` silently passing

trevyn:
Counterpart to https://github.com/mimblewimble/grin/pull/3632, fixes Windows `cargo test` to actually test instead of silently passing.

### DIFF
--- a/.ci/windows-release.yml
+++ b/.ci/windows-release.yml
@@ -6,10 +6,7 @@ steps:
       ROARING_ARCH: x86-64-v2
     displayName: Cargo Test All
     condition:  and(succeeded(), contains(variables['Build.SourceBranch'], 'refs/tags/'), eq(variables['CI_JOB'], 'release' ))
-  - script: 'cargo clean'
-    displayName: Cargo Clean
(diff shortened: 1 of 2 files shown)


.ci/windows-release.yml
```diff
@@ -1,22 +1,19 @@
...
+      LIBCLANG_PATH: C:\Program Files\LLVM\lib
+      LLVM_CONFIG_PATH: C:\Program Files\LLVM\bin\llvm-config
+      ROARING_ARCH: x86-64-v2
     displayName: Cargo Test All
     condition:  and(succeeded(), contains(variables['Build.SourceBranch'], 'refs/tags/'), eq(variables['CI_JOB'], 'release' ))
   - script: 'cargo clean'
     displayName: Cargo Clean
     condition:  and(succeeded(), contains(variables['Build.SourceBranch'], 'refs/tags/'), eq(variables['CI_JOB'], 'release' ))
-  - script: |
-      cargo clean
```
quentinlesceller, line 14: So no more cargo clean (different than in grin-wallet)?

trevyn: cargo clean was already here as a separate step above, but you're right that it should be consistent. Fixed.

quentinlesceller (approved):
Will merge once CI pass.

PR #604, opened 2021-04-07
Use `remove_dir_all` crate for more robust Windows CI

trevyn:
`std::fs::remove_dir_all` has serious issues on Windows: https://github.com/rust-lang/rust/issues/29497

@XAMPPRocky, who wrote and owns the `remove_dir_all` crate, is on the Rust release team (see https://www.rust-lang.org/governance/teams/release), notes that the crate is used in `cargo` and `rustup`, and suggests moving the crate's implementation into `std` (https://github.com/rust-lang/rust/issues/29497#issuecomment-573353391), so I think we can trust this dependency. Given that it improves a real issue on Windows, I think that until this is fixed in `std`, it would be wise to avoid `fs::remove_dir_all` wherever possible.


quentinlesceller:
Good idea though not a big fan of having yet another dependency...

trevyn:
`remove_dir_all 0.5.3` is already in the `Cargo.lock` as dependency of `tempfile`, would you prefer to use that version?

It's also only added in `dev-dependencies`.

Let's see if it actually works...

trevyn:
Ok, it doesn't look like there are any bug fixes between `0.5.3` and `0.7.0`: https://github.com/XAMPPRocky/remove_dir_all/commits/master

Shall I use `0.5.3` instead and replace the other test instances of `fs::remove_dir_all` with `remove_dir_all::remove_dir_all`, and revert the `Cargo.lock`?

quentinlesceller:
Probably better to use the 0.7.0 version then.

quentinlesceller (approved):

PR #605, opened 2021-04-12
show accept_fee_base in wallet config

deevope:
Currently accept_fee_base is hidden in the current grin-wallet. This PR allow newly created grin-wallet config to show this toml key.

### DIFF
--- a/api/src/owner_rpc.rs
+++ b/api/src/owner_rpc.rs
@@ -1054,7 +1054,8 @@ pub trait OwnerRpc {
 				"tls_certificate_file": null,
 				"tls_certificate_key": null,
 				"dark_background_color_scheme": null,
-				"keybase_notify_ttl": null
+				"keybase_notify_ttl": null,
+				"accept_fee_base": null
 			},
 			"logging_config": {
 				"log_to_stdout": false,
--- a/config/src/types.rs
+++ b/config/src/types.rs
@@ -74,7 +74,7 @@ impl Default for WalletConfig {
 			tls_certificate_key: None,
 			dark_background_color_scheme: Some(true),
 			keybase_notify_ttl: Some(1440),
-			accept_fee_base: None,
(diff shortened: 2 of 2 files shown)


phyro:
Thanks for opening the change! Could you explain a bit more what the change is e.g. what does it mean to show something in config? Where will this be shown?
Some tests seems to be failing

deevope:
@phyro This PR add the `accept_fee_base = 500_000` in the grin-wallet.toml config file for users that will generate a new config file with the grin-wallet version that will include this PR, it helps to choose the base_fee of the transaction you will send. Right now, if you go in your grin-wallet.toml file, you will not see this toml key because for some reasons it was disabled by default keybase://chat/grincoin.teams.wallet_dev#general/8562 

For the CI test, it failed for unrelated reasons to this PR. I triggered a new build and i guess this one will pass. Already had a similar unrelated issue with one of my previous PR...

```thread 'wallet_self_send' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 32, kind: Other, message: "The process cannot access the file because it is being used by another process." }', controller\tests\common\mod.rs:79:50```

GeneFerneau:
> thread 'wallet_self_send' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 32, kind: Other, message: "The process cannot access the file because it is being used by another process." }', controller\tests\common\mod.rs:79:50

Looks like some racy thing between multiple `controller` tests trying to access the same file. Maybe open a separate tracking issue? If you don't want to, I can.

GeneFerneau (approved):

phyro (approved):

phyro:
@quentinlesceller this one seems good to be merged

phyro:
ping @quentinlesceller @antiochp

deevope:
Minor commit, just noticed `accept_fee_base` was missing in the owner_rpc docs.

Issue #595, opened 2021-03-13
Getting the transaction list by the block height (Owner API)


davidtavarez:
I would like to filter the transaction list  by block height. Right now, for `retrieve_txs ` we have this:

```
{
  "jsonrpc": "2.0",
  "method": "retrieve_txs",
  "params": {
   "token": "token",
   "refresh_from_node": true,
   "tx_id": null,
   "tx_slate_id": null,
  },
  "id": 1
 }
```

I would like to also send `confirm_height` as a parameter, like this for example:

```
{
  "jsonrpc": "2.0",
  "method": "retrieve_txs",
  "params": {
   "token": "token",
   "refresh_from_node": true,
   "tx_id": null,
   "tx_slate_id": null,
   "confirm_height":  1127853,
  },
  "id": 1
 }
```

davidtavarez:
I can't see if the confirmed block height is stored: https://github.com/mimblewimble/grin-wallet/blob/master/libwallet/src/types.rs#L792 or returned https://github.com/mimblewimble/grin-wallet/blob/master/libwallet/src/internal/updater.rs#L93

Is there any workaround for this?

pkariz:
I think i know why exchanges might need this for the deposit flow (if they use SRS flow - afaik they all do that currently). The problem is that after the exchange does R part of SRS, there exists the final kernel excess (`X`), but from the slates (which you get from wallet-api calls) you can only get `Xs` and `Xr` (in hex compressed). Now you have 2 options:
1) you get `Xs` and `Xr`, calculate `X = Xs + Xr` and hex compress it (that seems painful to me and prone to errors in implementations)
2) you call `retrieve_txs(tx_slate_id=<...>)[0]['kernel_excess']` to get the excess (easy way)
The problem is that exchanges have many transactions and it would be good to limit the returned transactions by block height (it's easy to know for exchanges by which height to filter).

The current `TxLogEntry` doesn't seem to have the "confirmed" block stored

trevyn:
@davidtavarez If you're interested in digging a bit deeper on this one, I can personally offer a bounty of US$1,000 paid in BTC if you can manifest a PR for this.

Issue #362, opened 2020-03-09
Refine `Confirmation Time` column in `grin-wallet txs`


lehnberg:
### Steps to reproduce:
1. `grin wallet txs`
1. In this table, there's a column for "Confirmation time", providing a time stamp for when a transaction was confirmed.

### Current behaviour:
* The time stamp in "confirmation time" logs the time when _the wallet_ realised that the transaction was confirmed, not necessarily when the transaction actually was confirmed on chain.

### Expected behaviour:
* The column should display information of when the transaction in question was confirmed on chain, for example by either:
   * Providing the time stamp of the block that confirmed transaction; or
   * Renaming `Confirmation Time` to `Confirmed at height` and instead providing the block height that confirmed the transaction.

PR #606, opened 2021-04-15
Add the parameter confirmed_height to retrieve_tx owner_api method

deevope:
Improvement related to #595 and #362

- Retrieve transaction in function of their confirmed block height*
- add block height to the txs logs 
-  wallet cmd argument to list txs in function of block height

I suggest to fix before the #607 issue before merging this PR. If this issue is fixed, this PR would need some more code, if not it can be merged.


wiesche89:
Closing in favor of #802 , which provides an updated implementation.

PR #608, opened 2021-04-27
Added hw/error.rs, hw/apdu_types.rs and hw/types.rs

ndcroos:
Added a new directory 'hw' under the directory 'libwallet'.
Added the file 'error.rs', which contains the error definitions for the hardware ledger wallet on the host side.
Added the file 'apdu_types.rs', which contains the type definitions for the APDU packets that is used in the protocol to communicate with the ledger device.
Added the file 'types.rs', which contains miscellaneous used types.

### DIFF
--- a/libwallet/src/hw/apdu_types.rs
+++ b/libwallet/src/hw/apdu_types.rs
@@ -1,0 +1,36 @@
+use crate::error::*;
+use trait_async::trait_async;
+
+#[derive(Debug)]
+/// Commands follow the ISO/IEC 7816-4 smartcard protocol.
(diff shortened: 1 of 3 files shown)


phyro:
I wasn't following this too much. @ndcroos @quentinlesceller are these changes something we want to push to master right now? I'm not sure who would review this tbh, perhaps we could use Valdo's knowledge here (or maybe we want to do that later?)

ndcroos:
At the moment, it doesn't need to be pushed to the master; it's more a work in progress.
So maybe best wait and close for the time being?
I would also wait with the review.

PR #610, opened 2021-05-05
Save change outputs to wallet backend

GeneFerneau:
When building inputs, save change outputs to the wallet backend

Replaces no-op of saving to a function-local HashMap

### DIFF
--- a/libwallet/src/internal/selection.rs
+++ b/libwallet/src/internal/selection.rs
@@ -94,15 +94,25 @@ where
 		context.add_input(&input.key_id, &input.mmr_index, input.value);
 	}
 
-	let mut commits: HashMap<Identifier, Option<String>> = HashMap::new();
-
 	// Store change output(s) and cached commits
 	for (change_amount, id, mmr_index) in &change_amounts_derivations {
 		context.add_output(&id, &mmr_index, *change_amount);
-		commits.insert(
-			id.clone(),
-			wallet.calc_commit_for_cache(keychain_mask, *change_amount, &id)?,
-		);
+		let out_data = OutputData {
+			root_key_id: parent_key_id.clone(),
+			key_id: id.clone(),
+			n_child: 0,
+			commit: wallet.calc_commit_for_cache(keychain_mask, *change_amount, &id)?,
+			mmr_index: *mmr_index,
+			value: *change_amount,
+			status: OutputStatus::Unconfirmed,
+			height: current_height,
+			lock_height: 0,
+			is_coinbase: false,
+			tx_log_entry: None,
+		};
+		let mut batch = wallet.batch(keychain_mask)?;
+		batch.save(out_data)?;
+		batch.commit()?;
 	}
 
 	Ok(context)


libwallet/src/internal/selection.rs
```diff
@@ -94,15 +94,25 @@ where
...
+			value: *change_amount,
+			status: OutputStatus::Unconfirmed,
+			height: current_height,
+			lock_height: 0,
+			is_coinbase: false,
+			tx_log_entry: None,
+		};
+		let mut batch = wallet.batch(keychain_mask)?;
+		batch.save(out_data)?;
+		batch.commit()?;
```
phyro, line 115: If I understand it right, before we didn't do anything because as you said, it was a noop since we never used the hashmap. The `build_send_tx` seems to have quite a few consumers:
1. [late-locked finalize](https://github.com/mimblewimble/grin-wallet/blob/master/libwallet/src/api_impl/owner.rs#L787-L793)
2. `build_send_tx` is also used in [add_inputs_to_slate](https://github.com/mimblewimble/grin-wallet/blob/master/libwallet/src/internal/tx.rs#L142) which seems to be used in two cases. One is [init_send_tx which is not late-locked](https://github.com/mimblewimble/grin-wallet/blob/master/libwallet/src/internal/tx.rs#L142) and the other is [process_invoice_tx](https://github.com/mimblewimble/grin-wallet/blob/master/libwallet/src/api_impl/owner.rs#L668)

I suppose you hit some issue where your wallet didn't insert the "unconfirmed" commit when you expected it to. I'm trying to figure out through which flow this was an issue and whether this change could impact any other flows (to prevent adding the output twice or smth). Did you encounter this issue when going through any of the flows I listed above?

GeneFerneau: >  I suppose you hit some issue where your wallet didn't insert the "unconfirmed" commit when you expected it to.

Indeed, I couldn't figure out why the transaction reconstruction wasn't working right during atomic swap finalization. It was because the change outputs were never stored, causing the kernel offset to be incorrect.

DavidBurkett: @GeneFerneau The change outputs _are_ stored. They're in the context. The consumers of `build_send_tx` already commit that context to the database, so why can't you just read from that in order to calculate the kernel offset?

Can you point me to the code where transaction reconstruction fails due to this change output being missing?

DavidBurkett: Btw, if we do decide to keep this DB write here (I don't think we will), we should at least only have a single batch & commit, rather than a separate commit for each iteration of this for loop.

GeneFerneau: > Btw, if we do decide to keep this DB write here (I don't think we will), we should at least only have a single batch & commit, rather than a separate commit for each iteration of this for loop.

I am not sure how to do the single batch commit, since aquiring the wallet batch requires a mutable borrow, and the wallet is borrowed again inside the loop (which is why I decided to do the batch-per-loop-iteration).

> The consumers of build_send_tx already commit that context to the database, so why can't you just read from that in order to calculate the kernel offset?

The *current* consumers somehow manage saving change outputs, not sure exactly how. Maybe finding that out will remove the need to save the change outputs here. The private context is saved, but the change outputs are not. I only ran across this problem during atomic swap impl, so the problem doesn't exist outside that PR.

The current code does save the change outputs to a function-local HashMap, though (which is a no-op).

I can re-review my changes in the atomic swap impl, to see if I'm missing something when adding inputs to the swap transactions.

GeneFerneau: > Btw, if we do decide to keep this DB write here (I don't think we will), we should at least only have a single batch & commit, rather than a separate commit for each iteration of this for loop.

I am not sure how to do the single batch commit, since acquiring the wallet batch requires a mutable borrow, and the wallet is borrowed again inside the loop (which is why I decided to do the batch-per-loop-iteration). For example, I don't think doing a `wallet.batch(keychain_mask)?` before the loop, and a `batch.commit()?` after the loop will work.

> The consumers of build_send_tx already commit that context to the database, so why can't you just read from that in order to calculate the kernel offset?

The *current* consumers somehow manage saving change outputs, not sure exactly how. Maybe finding that out will remove the need to save the change outputs here. The private context is saved, but the change outputs are not. I only ran across this problem during atomic swap impl, so the problem doesn't exist outside that PR.

The current code does save the change outputs to a function-local HashMap, though (which is a no-op).

I can re-review my changes in the atomic swap impl, to see if I'm missing something when adding inputs to the swap transactions.

GeneFerneau:
Thanks @phyro and @DavidBurkett for the review, turns out I wasn't locking the outputs in the controller tests :facepalm: 

Apologies for not catching this sooner.

PR #613, opened 2021-05-12
Specify tor listen port in config

GeneFerneau:
Specify tor listen port as supplied in the onion service local address

Using port 80 requires root, so default to port 13415

### DIFF
--- a/impls/src/tor/config.rs
+++ b/impls/src/tor/config.rs
@@ -180,10 +180,23 @@ pub fn output_torrc(
 	props.add_item("SocksPort", socks_port);
 	props.add_item("DataDirectory", &tor_data_dir);
 
+	let listen_port = match wallet_listener_addr.find(|c| c == ':') {
+		Some(p) => wallet_listener_addr[p + 1..].parse::<u16>().map_err(|e| {
+			Error::from(ErrorKind::GenericError(format!(
+				"invalid onion listen port: {}",
+				e
+			)))
+		})?,
+		None => 13415,
+	};
+
 	for dir in service_dirs {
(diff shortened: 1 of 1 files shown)


deevope:
I don't think we should change the 80 port otherwise the sender must know the tor onion port of the receiver
which isn't possible to know right now with the slatepack address.

GeneFerneau:
> I don't think we should change the 80 port otherwise the sender must know the tor onion port of the receiver
which isn't possible to know right now with the slatepack address.

Yeah, I changed it from `80` because I was having an issue with `tor` not starting up. When I changed it to the default testnet port, it worked, but I think it may have been unrelated.

Regarding the sender not knowing the receiver port, we could pick a reasonable default not in the range that requires root privileges. Has the advantage of also being able to run an HTTP listener (clear web server), alongside the wallet (maybe someone has another website over `tor` on the same host?).

Picking `13415` was an oversight on my part, been doing a lot of stuff on testnet. Maybe we should pick the default listener port for each net type?

davidtavarez:
> Picking `13415` was an oversight on my part, been doing a lot of stuff on testnet. Maybe we should pick the default listener port for each net type?

Why the default port for each net type should be different?

GeneFerneau:
> Why the default port for each net type should be different?

It doesn't have to be, just a thought to match the differences in port number per net type on the local listener. A single default port (the one for mainnet?) could also work just as well.

DavidBurkett:
> Using port 80 requires root, so default to port 13415

No, this is port 80 of the hidden service listener, not the local machine's port 80. It's just a virtual port (https://community.torproject.org/onion-services/setup/#step-2:-configure-your-tor-onion-service), and should not require root at all.

PR #625, opened 2021-11-11
Update Cargo.lock from latest master, update age library to 0.7

yeastplume:
May have been due to recent changes in https://github.com/mimblewimble/grin/pull/3655, but wallet master was not compiling against latest grin master due to conflicts in the [age library](https://github.com/str4d/rage).

* Cargo update to update all crates
* Manually updated age crate to 0.7. Very annoyingly, the API for this has changed in such a way that it can no longer accept a raw x25519_dalek key to operate upon, and it took a while digging through source to come up with an inelegant fix whereby the keys are to be converted to bech32 strings before use.


libwallet/src/slatepack/types.rs
```diff
@@ -195,14 +195,16 @@ impl Slatepack {
 		b.copy_from_slice(&result[0..32]);
 
 		let x_dec_secret = StaticSecret::from(b);
-		let key = age::keys::SecretKey::X25519(x_dec_secret);
+		let x_dec_secret_bech32 =
+			bech32::encode("age-secret-key-", (&x_dec_secret).to_bytes().to_base32())?;
```
j01tz, line 199: Just want to double check that this bech32 string can be cast as an age `Identity` in the next line without calling `.to_uppercase()` on the bech32 string, as iirc age library expects this string to be uppercase.

j01tz (commented):
Once we confirm our age `Identity` is not case dependent, good to go. Thanks for cleaning this up.

libwallet/src/slatepack/types.rs
j01tz: After taking a closer look, this was changed to a lower case HRP so the code here is correct
https://github.com/str4d/rage/blob/456ce707f6db01f2f953c9e22c058964f3437795/age/src/x25519.rs#L20

j01tz (approved):

PR #630, opened 2021-12-11
View Wallet function (rewind_hash and scan rewind_hash)

deevope:
Features (api&cmdline): 
- get the rewind_hash of the wallet
- scan of the rewind_hash of a third-party wallet

Goal: 
Ensure the transparency (spending, receiving, balance etc) of a wallet that receive grin as donations. 
Allow everyone to scan the rewind_hash of the wallet shared publicly.

Cmd usage: 
get_rewind_hash: `grin-wallet rewind_hash`
scan_rewind_hash:  `grin-wallet scan_rewind_hash 2c95b24de492395934a8a345440ed0ebbe67ee0025b348712a927a91c7fe58e9`

PR #631, opened 2021-12-13
Update to latest master and ser&des fn arguments

deevope:
With the PR #630, I need to fetch actually the lastest master. But the test are failing due to the fact that the des&ser function are taking additional arguments with this new commit https://github.com/mimblewimble/grin/commit/63c65605bbe7a3ee4a8805c7f5aa7e7e153a15d5


yeastplume:
Thanks for this! 

I think CI is broken here, going to merge and check on master.

yeastplume (approved):

PR #636, opened 2022-01-06
Increase network read timeout when sending transactions

NicolasFlamel1:
These changes increase the network read timeout from 20 seconds to 5 minutes when sending transactions to avoid timing out too early when exchanging a slate with a device that takes longer than 20 seconds to generate a reply.

This will improve interoperability with wallets that support the [Grin dependent Ledger app](https://github.com/NicolasFlamel1/ledger-mimblewimble-coin) since that app takes approximately 90 seconds to generate a bulletproof on current Ledger hardware wallets.

### DIFF
--- a/impls/src/adapters/http.rs
+++ b/impls/src/adapters/http.rs
@@ -27,6 +27,7 @@ use crate::tor::config as tor_config;
 use crate::tor::process as tor_process;
 
 const TOR_CONFIG_PATH: &str = "tor/sender";
+const SEND_TX_TIMEOUT_DURATION_SECTIONS: u64 = 5 * 60;
 
 #[derive(Clone)]
 pub struct HttpSlateSender {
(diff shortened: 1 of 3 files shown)


yeastplume:
Thank you for looking into this. I agree that time timeout can be too short, particularly over the famously slow TOR network. However, I'm not sure the answer is to set it as high as 5 minutes by default in case of genuine timeouts.

Don't have a definitive answer here. Do you think it would be worth making this a configurable parameter?

ardocrat:
Reasonable value is 60sec, with ability to configure at `grin-wallet.toml`.

aglkm:
Grin-wallet is trying to send over Tor by default and increasing timeout value would lead to bad UX, because a slatepack address(tor) is not online in most cases, so that would be annoying to wait a long time.

Instead, a configurable parameter could be a good alternative to satisfy Ledger app users, who can be guided to increase the timeout.

Increasing it to 60s looks way high and that doesn't cover the issue reported by Nicolas.

ardocrat:
> Grin-wallet is trying to send over Tor by default and increasing timeout value would lead to bad UX, because a slatepack address(tor) is not online in most cases, so that would be annoying to wait a long time.

This PR is not about Tor sending, its about node-wallet communication.

60s is current timeout for Tor connection, whats can be normal, cause it can take some time to boostrap, otherwise we can decided Tor is not available. (Implemented at this PR https://github.com/mimblewimble/grin-wallet/pull/750)

Timeout for Tor connections also can be added to `grin-wallet.toml` config.

ardocrat:
Implemented at https://github.com/mimblewimble/grin-wallet/pull/765

PR #650, opened 2022-06-24
Show QR code for listener address

cliik:
Resolves https://github.com/mimblewimble/grin-wallet/issues/649.

To see it in action, compile and run `grin-wallet listen`. You will see your Slatepack address as usual, but now also with a scanable QR code:
```bash
$ grin-wallet listen
Password:
20220626 22:39:35.332 WARN grin_wallet_controller::controller - Starting Tor Hidden Service for API listener at address x5fnata4ge6fbj35hwojo2eynhxq5lgzr6e3q4lbd4unoxfmekbzw7qd, binding to 127.0.0.1:3415
20220626 22:39:37.720 WARN grin_wallet_controller::controller - Starting HTTP Foreign listener API server at 127.0.0.1:3415.
20220626 22:39:37.721 WARN grin_wallet_controller::controller - HTTP Foreign listener started.
20220626 22:39:37.722 WARN grin_wallet_controller::controller - Slatepack Address is: grin1ha9dqnquxy79pfma8kwfw6ycd8hsatxe37ymsutpru5dwh9vy2ps7xl420


   █▀▀▀▀▀█  █▄▄ ▀ █▄ █▄▀▀█ ▀▄█▄█ █▀▀▀▀▀█
   █ ███ █ █▀█▄██▄ █▄ █▀▀▄ █▄▀▄█ █ ███ █
   █ ▀▀▀ █ █ ▄ ▀▄▄ █ ▄▀█ █▀ ▀▀▄▄ █ ▀▀▀ █
   ▀▀▀▀▀▀▀ █ █▄█ ▀ █ ▀▄▀ ▀▄▀ ▀▄▀ ▀▀▀▀▀▀▀
   █▄█▀█▀▀▄ ▄█▀ ▀ █▄ ▀ ▀█▀ ▀█▀ ▀▄▀▀▀█▀
    ▀▄▄█▀▀▀▀▀▄▄▀   ▄ ▄ ▀▄██▀ ▄▀▀▀█ ▄▀ ▀▀
   ▀▄▀▄█▀▀ █▄▄█ ▄█ ██ ▄█▄▄█ ██▀██▀▄▀▄█▄▀
   █   ▀█▀▀█ ▀▀ ▀▄███▄  █▄▄▄ ▀█ ▄▄▀▄  ▀█
   ▄▀▀▄ █▀▀▄ ▄██ █▄▀▄▄▀▄█▄▀▄▀▄ ██▀▄▀█▀█▀
[...]

PR #652, opened 2022-06-30
Fix macOS CI tests

quentinlesceller:
Looks like the newest version of the runner does not include llvm anymore. This PR fixes the CI and unblocks #650.


yeastplume (approved):
LGTM

Issue #601, opened 2021-04-04
Ability to "send all" or "send max" with auto-calculation of fees


trevyn:
Feature request by @nthrow in #582. Something like:

```
grin-wallet send -e -m all
```

cliik:
I don't think the `-e` or `-m` options should be necessary for this feature.

I propose the following:
* amount may be `all` to send the entire wallet balance accounting for fees
* a confirmation prompt to confirm that the user really wishes to spend the entire wallet balance

I volunteer to implement this, if there's agreement on the feature request.

phyro:
Correct me if I'm wrong as I've never used this, but it seems we already have an option from this argument https://github.com/mimblewimble/grin-wallet/blob/master/src/bin/grin-wallet.yml#L120-L126

cliik:
Oh, perhaps I misunderstood that option. I thought `-s all` selected all outputs for a transaction, but did not necessarily compute the max amount you could spend less fees. i.e. you can send a transaction with `-s all` to someone, but you still need to specify the amount to send (vs receive back as change).

I think what @trevyn is requesting is a bit different. He is requesting a way to send the maximum amount to a recipient, accounting for fees. This is very similar to the `-s all` option, except there is no change output, and the wallet must calculate the transaction fees for the user.

I may be wrong though. Documentation for the `-s all` feature is slim, and I also have not used it myself.

phyro:
Thanks for clarifying, I think I now see what I missed. I wrongfully assumed the intent was to use the opportunity to sweep all outputs into a change output, but if you want to send max to someone else you have to specify the amount in which case you'd have to set the amount to `max_balance - fees(n_inputs, n_outputs, n_kernels)`. The user can't easily know the fee calculation. If we introduced `all` for the amount, we should be careful how we do that. If I recall correctly, the `SendTxArgs` is the CLI argument specification while we have `InitTxArgs` for the API. The question would be if this is a CLI option or also an option on the API. The former would mean you parse `SendTxArgs.amount` and compute the actual amount if `all` and do the api call on the computed amount. The latter would be a breaking change for the API. There's also something yeastplume mentioned to me the other day on keybase regarding transactions where you have no change outputs:
> like they need to scan kernels to check if the transaction completed, as opposed to waiting for a change output. So perhaps always having a change output would be cleaner, if we had a sweep strategy

I do wonder if introducing an option where the `send` command does not explicitly set the _exact_ amount is a good thing though.

cc: @yeastplume

[...]

PR #653, opened 2022-07-06
Implement 'max' value send

cliik:
This PR adds a special value 'max', which can be accepted as a spend amount. If this amount is specified, the wallet will calculate the maximum spendable amount (less fees) and send that amount. This is effectively a sweep.

PR is currently in draft state, as I need to test and think through edge cases some more.

Resolves https://github.com/mimblewimble/grin-wallet/issues/601

### DIFF
--- a/controller/src/command.rs
+++ b/controller/src/command.rs
@@ -16,7 +16,9 @@
 
 use crate::api::TLSConfig;
 use crate::apiwallet::{try_slatepack_sync_workflow, Owner};
+use crate::command::core::amount_to_hr_string;
 use crate::config::{TorConfig, WalletConfig, WALLET_CONFIG_FILE_NAME};
+use crate::core::libtx::tx_fee;
 use crate::core::{core, global};
 use crate::error::{Error, ErrorKind};
 use crate::impls::PathToSlatepack;
@@ -321,6 +323,7 @@ where
 #[derive(Clone)]
 pub struct SendArgs {
 	pub amount: u64,
+	pub max: bool,
 	pub minimum_confirmations: u64,
 	pub selection_strategy: String,
 	pub estimate_selection_strategies: bool,
@@ -341,7 +344,7 @@ pub fn send<L, C, K>(
 	owner_api: &mut Owner<L, C, K>,
 	keychain_mask: Option<&SecretKey>,
 	tor_config: Option<TorConfig>,
-	args: SendArgs,
+	mut args: SendArgs,
 	dark_scheme: bool,
 	test_mode: bool,
 ) -> Result<(), Error>
@@ -350,6 +353,33 @@ where
 	C: NodeClient + 'static,
 	K: keychain::Keychain + 'static,
 {
+	// If 'max' spend, calculate the 'amount' as the maximum spendable wallet balance, less fees.
+	if args.max && args.amount == 0 {
+		let current_height = owner_api.node_height(keychain_mask)?.height;
+		let utxos = owner_api.retrieve_outputs(None, false, false, None)?.1;
+		let spendable = utxos.iter().filter_map(|c| {
+			if c.output
+				.eligible_to_spend(current_height, args.minimum_confirmations)
+			{
+				Some(c.output.value)
+			} else {
+				None
+			}
+		});
+		let num_inputs = spendable.clone().count();
+		let spendable_amt = spendable.sum::<u64>();
+		let fee = tx_fee(num_inputs, 2, 1);
+		args.amount = spendable_amt.checked_sub(fee).ok_or(ErrorKind::LibWallet(
+			libwallet::ErrorKind::NotEnoughFunds {
+				available: spendable_amt,
+				available_disp: amount_to_hr_string(spendable_amt, false),
+				needed: spendable_amt + fee,
+				needed_disp: amount_to_hr_string(spendable_amt + fee, false),
+			},
+			"Wallet balance is less than fees.".to_string(),
+		))?;
+	}
+
 	let mut slate = Slate::blank(2, false);
 	controller::owner_single_use(None, keychain_mask, Some(owner_api), |api, m| {
 		if args.estimate_selection_strategies {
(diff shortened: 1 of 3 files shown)


controller/src/command.rs
```diff
@@ -350,6 +352,16 @@ where
 	C: NodeClient + 'static,
 	K: keychain::Keychain + 'static,
 {
+	// If 'max' spend, calculate the 'amount' as the maximum spendable wallet balance, less fees.
+	if args.max && args.amount == 0 {
+		let utxos = owner_api.retrieve_outputs(None, false, false, None)?.1;
+		let spendable = utxos.iter().filter_map(|c| match c.output.status {
```
phyro, line 358: here's a way to find outputs eligible to spend https://github.com/mimblewimble/grin-wallet/blob/master/libwallet/src/internal/selection.rs#L563-L570
Note that it also takes into account the 'account'. This is something I haven't thought about before. I'm not sure what should happen if you set `max` and also a specific wallet account. I'm guessing the user would expect to use max from this account? on the other hand, if the account wasn't specified, it defaults to the main account which might make it unclear that it would only use the max from the main account.

cliik: Perfect! I wondered if a method already existed for this.Thanks!

> I'm not sure what should happen if you set max and also a specific wallet account.

I think all spend commands are account specific, so IMO this one should be too. If the user specifies an account, this _should_ (it doesn't atm) spend max from that account. If the user doesn't specify an account, it should spend max from the default account. I will use the improved method you suggest, and be sure to use the account specified by the user (or the default if unspecified).

I realize I also need to add documentation for this & update the help dialog for the send command. I'll be sure to be clear about this there.

Thanks for the quick feedback!

cliik:
I've taken the time to test the following scenarios, and all worked as expected:
* normal send/receive functionality -- still works as before :heavy_check_mark:
* send max with zero coins in wallet -- failed gracefully with insufficient funds error :heavy_check_mark:
* send max with only unspendable coins in wallet -- failed gracefully with insufficient funds error :heavy_check_mark:
* send max with mix of spendable/unspendable -- correctly computed fee & spent available coins :heavy_check_mark:
* send max with only spendable coins -- correctly computed fee & spent entire balance with no change :heavy_check_mark:

I cannot think of any other cases to test and I feel comfortable marking this ready for review. cc @yeastplume

cliik:
@quentinlesceller the windows tests failed for a reason that appears to be unrelated to this feature:
```
thread 'slatepack_api' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 32, kind: Uncategorized, message: "The process cannot access the file because it is being used by another process." }', controller\tests\common\mod.rs:79:50
```
Can you trigger the workflow to run again, to see if it was a transient failure?

cliik:
> Can you trigger the workflow to run again, to see if it was a transient failure?

I decided to force-push to trigger a rerun

davidtavarez:
I was looking for this feature recently! @cliik would you mind to write some tests for this?

yeastplume:
The implementation looks good and seems to work well. The wallet has extensive tests to ensure as much is covered as possible and it keeps working well, so the next step would be to automate all of the tests you've outlined above. Here's the general process:

* Start an integration test by copying one of the existing test files in `controller\src\tests`. These simulate wallet-to-wallet and wallet-to-node communication so you should be able to add tests to cover all of the new cases you outlined above and check the results.
* Check the API tests to make sure the samples don't need adjustment. In this case, `send_args` aren't really used in the doctest examples, so I don't think there's much further to do here as long as the `InitTxArgs` field is documented.

When adding a feature, I usually start by developing one of the tests in the `controller\src\tests` director to exercise it, which also saves a lot of development time cause I don't need to manually run wallets and check the results. Also easy to add new test cases there to catch explicit conditions if a bug or issue comes up.

cliik:
@davidtavarez @yeastplume I've spent several hours trying to get automated tests working for the CLI, but hitting a roadblock. Can you point me to any tests that exercise the wallet commands, instead of the wallet API? So far all I see are tests for the wallet API, which is not useful for this feature, because the change does not exist in the API methods.

Alternatively, I could re-implement this feature in the API, instead of in the CLI command processing. I didn't want to make an API change, because this feature felt like something only intended for CLI users. API users can easily calculate the appropriate send amount, and I didn't want to suggest an API change where it isn't warranted... although perhaps this was the wrong choice.

yeastplume:
I'm of two minds as well as to whether this should be in the API, I think it may be fine to leave it as is for now and we can revisit later. If it's just adding an argument to the send args it shouldn't break any existing APIs, but it would still need more thought and consideration than placing it where you have it now.

For testing at the CLI level, look in the top level `src\tests\cmd_line_basic.rs` for examples

davidtavarez:
@cliik @yeastplume what do you of instead of adding a parameter the API can receive `amount` as `null` and that will means to use `max`; that's how others API are doing this. Just an idea.

cliik:
@yeastplume ahhthanks! Idk why it was so hard for me to find those CLI tests. I'll take a look later/tomorrow.

@davidtavarez in general I prefer to be explicit. If someone accidentally omits/nulls an argunent, the wallet shouldn't proceed to empty their coins.

If I was to move this to the API (which I'm starting to become in favor of), it might make more sense to change InitTxArgs.amount to an enum with two variants. One variant being the usual specified amount, and another variant being the max spend option. This would be clean & simple, and also opens the door for other variants in the future (i.e. if we wanted a variant to spend X% balance or something like that)... I dont think we should feature creep here, lol, but future extensibility is an additional benefit worth mentioning.

cliik:
Ok, I've thought about this feature some more and I'm not happy with my original approach. After brainstorming a few better solutions, I also don't like using an enum with special variants for the amount. That would have rippling consequences in several other parts of the codebase, and the changes would be messy at best.

I'm going to close this PR for now, and discuss the new approach I have in mind in the original issue:
https://github.com/mimblewimble/grin-wallet/issues/601

cliik:
This PR has been replaced by a better approach (IMO): https://github.com/mimblewimble/grin-wallet/pull/657

PR #654, opened 2022-07-06
Replace `failure` with `thiserror`

yeastplume:
Wallet counterpart for https://github.com/mimblewimble/grin/pull/3728, note that the node PR will need to be merged before this will compile. 

All should be as before, except `thiserror` doesn't support explicit backtraces so these have been removed from test output where they appeared.

Review should again just check to ensure I haven't accidentally changed any logic.


phyro (approved):
Great job! :+1:

I went through the changes and didn't notice any logical mistakes. I also made some minor calls to CLI/API that returned an error and got the same response :+1:

yeastplume:
All good, many thanks for reviewing!

PR #656, opened 2022-07-10
Do not prompt for password to start owner API

davidtavarez:
It makes no sense to ask for the password if one only wants to initialize the APIs. Asking for the password makes things unnecessarily difficult and insecure in case I, as a system administrator, want to have the APIs running as a system service.

### DIFF
--- a/src/cmd/wallet_args.rs
+++ b/src/cmd/wallet_args.rs
@@ -1018,7 +1018,7 @@ where
 		("init", Some(_)) => open_wallet = false,
 		("recover", _) => open_wallet = false,
 		("cli", _) => open_wallet = false,
-		("owner_api", _) => open_wallet = false
+		("owner_api", _) => open_wallet = false,
 		_ => {}
 	}
 
--- a/src/cmd/wallet_args.rs
+++ b/src/cmd/wallet_args.rs
@@ -1019,7 +1019,7 @@ where
 		("recover", _) => open_wallet = false,
 		("cli", _) => open_wallet = false,
 		("owner_api", _) => {
-			// If wallet exists and password argument is present, open it. Otherwise, that's fine too.
+			// If wallet exists and password is present then open it. Otherwise, that's fine too.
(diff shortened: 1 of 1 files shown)


phyro:
@yeastplume I'm not familiar with this part. The [wallet lifecycle RFC](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0004-full-wallet-lifecycle.md) says
> 'Open' wallets store their in-memory seeds XORed against the token, which is temporarily XORed against the supplied token during each request to reproduce the master seed.

Is opening a wallet here by design or can we remove it?

davidtavarez:
> @yeastplume I'm not familiar with this part. The [wallet lifecycle RFC](https://github.com/mimblewimble/grin-rfcs/blob/master/text/0004-full-wallet-lifecycle.md) says
> 
> > 'Open' wallets store their in-memory seeds XORed against the token, which is temporarily XORed against the supplied token during each request to reproduce the master seed.
> 
> Is opening a wallet here by design or can we remove it?

It can be remove it. When the `open_wallet` method is called, the token is generated. If I want to use the APIs I should not be asked to enter the password wallet and then anyways get the token via que API using `open_wallet` method.

My guess is that since the token can be null while using [init_secure_api](https://github.com/mimblewimble/grin-wallet/blob/master/tests/data/v3_reqs/init_send_tx.req.json) one could call `init_secure_api` directly without calling `open_wallet` because the wallet is already open.

What changes this PR is that the wallet will not open unless someone passes the password directly while setting the `owner_api` flag.

What I want to achieve is to be able to configure a local service that runs the API. In case the instance running the service is restarted, the service could be started automatically after starting the instance without human intervention.

yeastplume:
I think this is fine, the password check is likely left over from before the implementation of the lifecycle API and tokens, when there was no concept of opening or closing wallets, just the single 'open wallet' on the listener.

yeastplume (approved):

Issue #580, opened 2021-01-16
Outdated code and docs


DavidBurkett:
"Method" is no longer supported to the best of my knowledge. 
https://github.com/mimblewimble/grin-wallet/blob/429db61bb7ef42f009abb8886fd90221301dfc74/src/bin/grin-wallet.yml#L59-L67

Keybase adapter still exists and should be removed:
https://github.com/mimblewimble/grin-wallet/blob/7f5c010fed2f9177e0e24cc3eaa7b79be112390f/impls/src/adapters/keybase.rs

Wiki doesn't cover slatepacks:
https://github.com/mimblewimble/docs/wiki/How-to-use-the-Grin-wallet

cliik:
I can't speak to the first two items, but I also have a question about the third:

Slatepacks are covered in [docs.grin.mw/wiki](https://docs.grin.mw/wiki/transactions/slatepack/), but this makes me wonder why we have two competing wiki's with different information. Should we consider obsoleting/removing the GitHub wiki in favor of the new `docs.grin.mw/wiki`? I'm willing to volunteer some time to help that migration, but I don't have enough context to decide what should stay or go.

cliik:
> I'd agree the duplicate entries should definitely go away.

I'd go a step further and say the entire duplicate wiki should go away, to prevent these sort of inconsistencies from arising again. Of course migrating wikis is no small effort, so I don't blame anyone for the current situation, but if we are going to address issues like this, it might be best to address the root cause by wholesale migrating the old wiki to the new location.

I'm not sure who would need to approve such a move (if it even would get approved), but if so, I am willing to volunteer some time to the effort. Clearly someone put a lot of thought into the old and new wikis, so probably best to consult those people. Do you know who to ask about this?

phyro:
I very much agree the optimal is to have a single reliable source in this case. Perhaps one thing to note is that maybe there are links to wiki page from articles in 2019 or similar e.g. to page Grin for bitcoiners or something. I'd say the best place to ask is to ask the community on the [forum](https://forum.grin.mw/). Someone that made great work on the documentation goes under Paouky on the forum. He hasn't been around for some time though.

ardocrat:
Not actual

PR #658, opened 2022-07-22
Remove method from grin-wallet.yml and keybase.rs from adapters

s7500:
This PR solves part of the [issue](https://github.com/mimblewimble/grin-wallet/issues/580)
Doesn't break consensus
No additional unit tests needed
All tests passed


phyro:
Thanks for the contribution @s7500 . I'll take a look into it a bit more later when I get to my PC. One thing that comes to mind is that removing keybase adapter would also require removing `keybase_notify_ttl` config option.

s7500:
Thanks @phyro, that's a good note!
I also removed it in the new commit

s7500:
Sorry, my mistake
I should trust my ide less

yeastplume:
Looks good, thanks for this. I remember it being annoying having to maintain that unused keybase adapter, so happy to have it removed from the code.

yeastplume (approved):

Issue #659, opened 2022-07-25
Should retrieve_txs() API allow for constrained response size?


cliik:
**Is your feature request related to a problem? Please describe.**
Yes. Retrieving TXs for a wallet with a large transaction history, may result in an unknowably large response. This could cause performance or UX issues for users with lots of transactions. Super users (or exchanges) with thousands or maybe millions of txs one day may have a problem with this API as it exists today.

**Describe the solution you'd like**
I think the API should allow the requester to constrain the response to a limited number of TXs.

**Describe alternatives you've considered**
A simple alternative, would be to add `count` and `start_at` args to the `retrieve_txs()` method. In addition to constraining the response to a limited number of results, this will also facilitate pagination. Pagination may be the best solution, when grin super-users or exchanges begin to stress the limits of a wallet with 10Ks/100Ks/1Ms of transactions.

**Additional context**
Constraints like this will also be valuable/necessary for wallets running in resource limited environments (RaspPi, hardware wallet, etc).

I volunteer to do this work if there is community buy-in around the idea. This would be a breaking change to the API, so it would be nice to coordinate this change with other breaking changes (e.g. https://github.com/mimblewimble/grin-wallet/pull/657).

cliik:
FYI, I've opened a CLI-only PR to address the more immediate UX issue I am experiencing (too many TXs to view in my ssh window!): https://github.com/mimblewimble/grin-wallet/pull/660. That PR does not address the forward-looking needs of the API though.

yeastplume:
Yes, this is going to be required for the wallet gui work, as we'll eventually need some kind of paging mechanism, so would be quite happy to see this. Again, I don't think it needs to be breaking as we're perfectly able to support optional arguments and flags throughout.

'Super Users' or exchanges need to be better supported overall, and one idea is to refactor the backend some to allow for sql databases as opposed to lmdb (which is a much larger project).

If you'd like to discuss what's going on in the gui a bit more I'd be happy to. There's likely to be a few things that fall out of it that turn out to be missing or really really nice to have, this being one.

wiesche89:
implemented: #660

PR #660, opened 2022-07-25
Add option to limit number of TXs shown

cliik:
This PR adds a `--count` option to the `txs` wallet command, to restrict the number of TXs shown. I find this very helpful when working in wallets with many transactions. Example usage:

```
# Display last 10 transactions
$ grin-wallet txs --count 10
```

This is a CLI change only, which I think is useful today. As I outline in https://github.com/mimblewimble/grin-wallet/issues/659, it might also be worth modifying the API to allow constrained queries to the wallet.

### DIFF
--- a/controller/src/command.rs
+++ b/controller/src/command.rs
@@ -1093,6 +1093,7 @@ where
 pub struct TxsArgs {
 	pub id: Option<u32>,
 	pub tx_slate_id: Option<Uuid>,
+	pub count: Option<u32>,
 }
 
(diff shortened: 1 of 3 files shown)


yeastplume:
Looks good from a CLI perspective, and the ability to page transaction requests is something that's likely to come up very soon during gui work: https://github.com/mimblewimble/grin-gui, will respond in the other thread

yeastplume (approved):

PR #661, opened 2022-07-28
Remove grin lib re-exports from util crate

yeastplume:
For convenience, the previous version of the code imported all grin libraries into the `util` crate and re-exported them for the rest of the wallet crates. This was done so that changes to versions or branches could be made in one place.

This is fine if the wallet is the only thing being compiled, but as discovered when working on the gui, importing libs for both `grin` and `grin-wallet` causes namespace conflicts.

The only solution is to ensure all grin crates are referenced by their original path, which means that the re-export needs to be removed and the grin libs imported into the wallet crate-by-crate. This unfortunately also means that we now need to update versioning info for builds or local development in each and every crate except for just the util crate.

It would be very useful to figure out a way to make changes for development/build/releases in a single place using whatever tools cargo provides, but that will have to be a future endeavour.


Cargo.toml
```diff
@@ -35,9 +35,38 @@ grin_wallet_impls = { path = "./impls", version = "5.2.0-alpha.1" }
...
+
+# For beta release
+
+# grin_core = { git = "https://github.com/mimblewimble/grin", tag = "v5.0.0-beta.2"}
+# grin_keychain = { git = "https://github.com/mimblewimble/grin", tag = "v5.0.0-beta.2" }
+# grin_util = { git = "https://github.com/mimblewimble/grin", tag = "v5.0.0-beta.2" }
+# grin_api = { git = "https://github.com/mimblewimble/grin", tag = "v5.0.0-beta.2" }
+
+# For bleeding edge
+ grin_core = { git = "https://github.com/mimblewimble/grin", branch = "master" }
```
phyro, line 57: nitpick: there are spaces/indentation that could be removed in this and other toml files

phyro (approved):
The nature of this one is such that it would result in compile errors if something was off. The changes look good :+1:

yeastplume:
Yes, agreed. Thanks for taking a look!

Issue #673, opened 2023-03-10
v5.1.0 segmentation fault when running account or info commands 


ardocrat:
I can not see my account balance, every time its crashing with segmentation fault (core dumped).

**To Reproduce**
1. Run `./grin-wallet account`
```
____ Wallet Accounts ____

zsh: segmentation fault (core dumped)  ./grin-wallet account
```

1. Run `./grin-wallet info`
```
...

WARN grin_wallet_libwallet::api_impl::owner_updater - Scanning Complete

____ Wallet Summary Info - Account 'default' as of height *height_number* ____

zsh: segmentation fault (core dumped)  ./grin-wallet info
```

There is nothing about error I can see at logs.

I have installed:
```
rustc 1.67.1
cargo 1.67.1
```

ardocrat:
Tried with grin node `v5.2.0-alpha.2`.
`v.5.1.2` is not building, described here: https://github.com/mimblewimble/grin/issues/3742

ardocrat:
Error is `attempt to add with overflow`

Debug builld showed me this:
```
ERROR grin_util::logger - 
thread 'main' panicked at 'attempt to add with overflow': /rustc/library/core/src/iter/traits/accum.rs:141   0: grin_util::logger::send_panic_to_log::{{closure}}
             at /home/user/.cargo/registry/src/github.com/grin_util-5.1.2/src/logger.rs:316:19
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/library/alloc/src/boxed.rs:2032:9
      std::panicking::rust_panic_with_hook
             at /rustc/library/std/src/panicking.rs:692:13
   2: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/library/std/src/panicking.rs:577:13
   3: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/library/std/src/sys_common/backtrace.rs:137:18
   4: rust_begin_unwind
             at /rustc/library/std/src/panicking.rs:575:5
   5: core::panicking::panic_fmt
             at /rustc/library/core/src/panicking.rs:64:14
   6: core::panicking::panic
             at /rustc/library/core/src/panicking.rs:111:5
   7: <usize as core::iter::traits::accum::Sum>::sum::{{closure}}
             at /rustc/library/core/src/iter/traits/accum.rs:45:28
   8: core::iter::adapters::map::map_fold::{{closure}}
[...]

PR #674, opened 2023-03-13
Manual update and verification of prettytable-rs update

yeastplume:
* Verified issue in #673 on running installation
* Had cursory look at dependency updates involved in bumping the version of prettytable-rs
* Manually implement #668


quentinlesceller (approved):
Just tested and looking good with grin 5.1.2

yeastplume:
thanks for verifying!

PR #675, opened 2023-03-20
Simple contracts restructured v3

yeastplume:
Merging Phyro's work to date into an new `contracts` branch for further discussion and consideration. Note this is not planned for any release as of yet.


yeastplume:
Branch is experimental, so not worried about tests passing at the moment.

PR #685, opened 2023-07-25
Update sysinfo crate to 0.29.6

yeastplume:
Due to changes in rust, compilation was failing on windows and we need to update the sysinfo crate used by tor process handling to the latest. This also comes with API changes, namely that the `Process` struct can't be instantiated directly and needs to be searched in the `System` struct. 

Changes now needs to be sanity tested on all platforms to ensure tor processes still start and terminate as expected.

Also contains updates to remove deprecation warnings from chrono related updates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.17.0"
+version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
+checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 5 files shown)


yeastplume:
Verified tor listener on windows and linux, Mac probably okay but can't check.

quentinlesceller:
For some reasons I'm getting a segfault when running this in release mode only on macOS arm64 (m1). Debug mode runs fine.
```
❯ ./target/release/grin-wallet listen
Password:
zsh: segmentation fault  ./target/release/grin-wallet listen
```

quentinlesceller:
This:
```
diff --git a/impls/src/tor/process.rs b/impls/src/tor/process.rs
index ed6627d..f964326 100644
--- a/impls/src/tor/process.rs
+++ b/impls/src/tor/process.rs
@@ -101,7 +101,7 @@ impl TorProcess {
                        working_dir: None,
                        stdout: None,
                        process: None,
-                       sys: System::new_all(),
+                       sys: System::new(),
                }
        }
 
 ```
 
 Seems to fix the segfault.

yeastplume:
Thanks! No, looks like new_all isn't needed here. Tested linux + windows again with changes.

yeastplume:
That single failing windows test is driving me insane. I'm just going to merge this and deal with that separately.

PR #686, opened 2023-07-25
Update tokio to v1 and reqwest to v0.11

quentinlesceller:
Based on the build issue we currently have on Windows. This PR update:
- tokio from 0.2 to 1.29
- reqwest from 0.10 to 0.11.

~~I'm expecting that some tests might not pass as this is a major upgrade.~~ Locally all tests are passing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -204,6 +204,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
+name = "base64"
+version = "0.21.2"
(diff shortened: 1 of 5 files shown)


yeastplume:
Yeah, this is likely to be a major upgrade that will need attention to what's changed as well as quite a bit of testing, let's see what errors it comes back with. Note we don't need this right now cause https://github.com/mimblewimble/grin-wallet/pull/685 fixes our current issues, but we likely want to do this soon, perhaps we can do a release soon then look at this right after?

quentinlesceller:
Yes absolutely. And indeed looks like #685 solved the CI issue. Agree to pause this until next release.

ardocrat:
tokio and reqwest were updated at https://github.com/mimblewimble/grin-wallet/pull/750

PR #688, opened 2023-07-25
Replace Azure Pipelines with Github Actions

quentinlesceller:
As titled. This PR replaces Azure Pipelines (AP) with Github Actions (GA).
Pros:
- GA does not require a microsoft account.
- GA is way more simple and robust.
- GA is well integrated with Github.

Cons: ?

As of today it does exactly the same as what we have:
- Tests all linux packages separately.
- Runs `cargo test --release --all` on macOS and Windows.
- Handles release including binaries, changelog and shasum. Example here https://github.com/quentinlesceller/grin-wallet/releases/tag/v5.2.0-test.1
- One caveat is that we can produce a release even if the tests are not passing. But that would be extremely strange to push a tag if our tests are not working.

To test:
- [x] ~~Windows Release (Waiting for #685 to be merged).~~ Tested and working.

If this is merged I intend to switch grin repo to Github Actions as well.

Also binaries arch are renamed to x86_64 instead of amd64 to avoid confusion with arm64.


quentinlesceller:
Is there anything else that you'd like in the build infrastructure?

yeastplume:
Let me have a think and come back to you - thanks!

yeastplume:
Are we okay to merge this? Would like to give it a go with a beta release

quentinlesceller:
Sure. Let's do this.

PR #697, opened 2023-10-17
[Contracts] Add self spend transaction state

yeastplume:
* Adds an explicit `SelfSpend` transaction state, at the moment only set during contracts workflow.
* Sets the type on a transaction during contract workflow when the number of participants in the slate is set to 1

Note this would have a lot of extra areas that needs to be addressed were it used in the legacy wallet workflow. This included on the contracts branch only for experimental purposes and reflected only in the gui project for now.


yeastplume:
Think this is okay now, mergring to contracts.

PR #702, opened 2024-02-22
Fixes to self-spend cancellation

yeastplume:
Uncovered via https://github.com/mimblewimble/grin-gui/issues/79, the logic for cancel was failing when attempting
to cancel a legacy self-spend (i.e. when there are two transaction log entries for the same transaction uuid)

* Updates cancel function to not abort when the number of transaction log entries for a uuid != 1
* Call cancel logic for all transaction log entries with given uuid
* Add test case covering cancelling self-spends by transaction uuid


yeastplume:
> @yeastplume the diff on this one shows only one test file while description hints some other changes should be present as well. Is it missing something?

Whoops yes :D , creating another PR with missing changes now.

PR #712, opened 2024-05-20
Versioning to 5.4.0-contracts.0

yeastplume:
* Include latest grin master
* Update versioning label to include contracts denotation


Anynomouss:
🚀 Does this version of _grin-wallet_ also include the new bi-directional payment proofs?

PR #715, opened 2024-06-02
Store retrieved kernel height at TxLogEntry structure

ardocrat:
Useful to store transaction kernel block height at `grin_wallet_libwallet::TxLogEntry` to calculate left confirmations amount and to know when transaction was exactly included into block without extra API calls from outside, now we are just marking tx as confirmed with current time at `update_txs_via_kernel` from `grin_wallet_libwallet::owner::update_wallet_state`.


ardocrat:
Makes no sense to request kernel height here, cause its always 0 for unconfirmed txs.

Issue #728, opened 2024-12-03
A minor issue with using "-h" option to run the wallet in current directory


aglkm:
When you are creating a wallet in the current directory using `grin-wallet init -h` command, you will end up with errors saying that wallet can't contact a node.
The reason is that wallet creates API secrets paths in grin-wallet.toml file pointing to the current directory, not the actual/correct paths of node's api secrets location.

It would be nice if wallet could auto-detect the correct paths, if not possible then just print a simple message, saying that you need to correct API secrets paths, e.g.:
```./grin-wallet --testnet init -h
20241203 22:44:37.332 ERROR grin_wallet_impls::node_clients::http - Error calling get_version: ResponseError error: Cannot parse response
20241203 22:44:37.332 ERROR grin_wallet_impls::node_clients::http - Unable to contact Node to get version info: Client Callback Error: Error calling get_version: ResponseError error: Cannot parse response
**WARNING: YOU NEED TO CORRECT API SECRET PATHS IN GRIN-WALLET.TOML FILE.**
Please enter a password for your new wallet
Password: 
Confirm Password: 
```

Anynomouss:
I just want to mention that this issue of not correctly detecting (or creating) the API secret when using -h or -t is quite a high priority bug in my opinion, since it blocks anyone from testing using multiple wallets on a single machine - unless they would run a docker for each wallet instance.

I am not sure even what exactly the root issue is here. 
For example copying the .api_secret secret from my default wallet directory to my custom top level directory does not make the wallet able to connect to the node API. Also the grin-wallet.toml contains lines for  .foreign_api_secret and .owner_api_secret but none for .api_secret. Which secret is exactly needed by grin-wallet to connect to the node?

aglkm:
This issue should not block from using multiple wallets.

The secrets inside wallet directory are for wallet API.
The secrets inside node directory are for node API.

Don't copy the secrets, but instead correct `node_api_secret_path` inside `grin-wallet.toml` file for each wallet you are using.

`node_api_secret_path` should point to node's Foreign API secret.

Node API secrets:
`.api_secret` - Node Full API access. ( ? )
`.owner_api_secret` - Node Owner API.
`.foreign_api_secret` - Node Foreign API.

Wallet API secrets:
`.owner_api_secret` - Wallet Owner API.
`.foreign_api_secret` - Wallet Foreign API.

Btw, I don't have `.api_secret` inside wallet directory. You copied it from node's directory?

[...]

Issue #478, opened 2020-07-16
Config file generation


reagentoo:
It would be nice if the grin-wallet can generate grin-wallet.toml to the current directory (without any other data).
It's similar how grin-server can do it:
```
$ grin server config
grin-server.toml file configured and created in current directory
$ ls grin*
grin-server.toml
```
This is useful for creating packages for different linux distributions and for being able to run with different init systems. (Systemd / OpenRC).

Let me describe more details. To start the application by the initialization system, you must bring the file storage paths to the following:
```
/etc/grin/grin-miner.toml
/etc/grin/grin-server.toml
/etc/grin/grin-wallet.toml

/usr/bin/grin
/usr/bin/grin-miner
/usr/bin/grin-wallet

/var/lib/grin/chain_data
/var/lib/grin/wallet_data

/var/lib/grin/.api_secret
/var/lib/grin/.foreign_api_secret
/var/lib/grin/.owner_api_secret

/var/log/grin/grin-miner.log
/var/log/grin/grin-server.log
/var/log/grin/grin-wallet.log
```

/etc/grin/*.toml should be installed by the package manager. But grin-wallet doesn't provide generation of the config file.

reagentoo:
Also grin-wallet does not process configuration files correctly. It is not possible to move grin-wallet.toml from working directory.
.api_secret and .owner_api_secret are always recreated in the same directory of grin-wallet.toml.

Anynomouss:
Since I already was working on everything config related, I also checked this issue.
The problem is that in your config, the main fields are not preceded by ##, the  migration function cannot deal with this:
Your file:

```
#########################################
LOGGING CONFIGURATION
#########################################
```
Expected by migrate_comments

```
#########################################
## LOGGING CONFIGURATION
#########################################
```
Also your old config contains two types the filed `chain_type`  causing an error. 
Regarding letting the wallet generate a config, is that  useful? A user can simply generate a new wallet to generate a template in another directory.

tromp:
Thanks for resolving this, @Anynomouss

Anynomouss:
[...]

PR #731, opened 2025-02-14
Fix for many wallet config errors

Anynomouss:
# Previous behavior:
wallet.rs creates config with updated _.api_secret_ _.foreign_api_secret_ and _.owner_api_secret_
wallet arguments `-h (here)` ad `-t (top_dir)` updated the paths in the config, including _.api_secret_ which was wrongly set to top_dir or in case of -h to current directory. Secondly, later on grin_wallet_controller::command::init is called , even though the config file is passed via _args_, a new config file is created.

# New behavior
All wallet config creation happens purely in wallet.rs using functions defined in config.rs.
The node path is determined flexibly, first checking top_dir if _.api_secret_ exist there, otherwise assuming home/.grin
Paths set by top_dir are relative and not absolute. Meaning the user can move around the folders that contains _grin_, _grin-wallet_ and its data as long as _grin_ binary does not move relative to the grin-wallet data dir. An alternative would be to make the path absolute.

# Issues that are fixed (or related).  Quite a few issue can be closed.
https://github.com/mimblewimble/grin-wallet/issues/478 [related, not fixed]
https://github.com/mimblewimble/grin-wallet/issues/728 [fixed]
https://github.com/mimblewimble/grin/issues/3803 [fixed]
https://github.com/mimblewimble/grin/issues/3394 [fixed]
https://github.com/mimblewimble/grin/pull/3420 [Abandoned Pull request, my pull replace this one]
https://github.com/mimblewimble/grin/issues/3002 [Partly fixed, not default values part]
https://github.com/mimblewimble/grin/issues/2300 [Could be closed, not dependent on my pull]

Note
The wallet firsts checks a) current directory for config, b) top_dir for config, then c) assumes its in home/.grin.
Option a) suppresses generation of a config file when running grin-wallet.exe init, not sure if that is intentional behavior or that the config should be crated but the one in current dir should be loaded.


Anynomouss:
I have been thinking, perhaps this pull might need a bit of reworking/discussing @yeastplume.

1) If there is a config in the current directory, _grin-wallet_ loads the path and the config file from the current dir, even when running `-init`.
Would it not be more elegant, to load config from current dir and use it as a template for the newly created config file? 
What I mean is to I) Load the config in the current directory as an object with all its particular settings and use it as template II) update that config object by setting the API ad data paths, wallet dir and 3) writing it to the wallet data dir as a new cofig file that did not use the default settings but the settings from the config file in the current directory. 

2) In this pull request I moved the writing of the config file from _commad.rs_ to _wallet.rs_. The downside is that with a config in the current dir, no new config is created. Only after reworking the config creation I found that the config as object is passed from wallet.rs to the owner API (although it is ignored when running `-init`). In contrast the config path is not passed along as argument and imported as global variable. What would be more elegant is to add the config_path as a variable to the config object, do the updating in wallet.rs and let the writing be done by command.rs. The only possible downside is that the config contains now the user path as part of the cofig_path which might be considered somewhat sensitive data.

aglkm:
I can't compile with this PR:
```
error[E0061]: this method takes 2 arguments but 1 argument was supplied
   --> impls/src/lifecycle/default.rs:143:18
    |
143 |         default_config.update_paths(&abs_path);
    |                        ^^^^^^^^^^^^----------- an argument of type `&PathBuf` is missing
    |
note: method defined here
   --> /home/user/grin/grin-wallet-sources/grin_wallet_pr731/grin-wallet/config/src/config.rs:332:9
    |
332 |     pub fn update_paths(&mut self, wallet_home: &PathBuf, node_home: &PathBuf) {
    |            ^^^^^^^^^^^^
help: provide the argument
    |
143 |         default_config.update_paths(&abs_path, /* &PathBuf */);
    |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

Anynomouss:
@aglkm That can very well be the case, I reverted some code, sloppy that I pulled that to main and have worked on it a bit since.
In any case this pull needs some work and is not ready for review. I either have the config by wallet.rs working updating the paths and writing the config file, but the user gets a warning because owner api does recognize the config is already created and tries to set its own config with default values. I need to dig a bit deeper to see how I can have all of them working nicely together without any conflicts and unneeded error messages tot he user.

Anynomouss:
The pull is ready for review. 
_Default.rs_ can still be improved, it does not properly work on its on its own with `-top-dir `and `init -h`, but because the config loading and updating now happens in _wallet.rs_, and `defaults.rs` only writes the config  it is not a problem. If no config is provided defaults creates its own config. 

**Tested and works for:**
a) grin-wallet.exe init
b) grin-wallet.exe init -h
c) grin-wallet.exe -t my_top_dir init . Replaced "/" with "\\" to keep the uses of slashes uniform in the config.
d) grin-wallet.exe init (providing a template config in the current directory)

yeastplume:
Changes look okay so far, but there are still quite a few test failures, see log from runs above

Anynomouss:
I will have a look at it. I did not yet read the Rust Book chapter on tests, so I simply forgot to run *cargo test* and adjust existing tests. I expect it to be little work.

Anynomouss:
@yeastplume Works now, at least on Windows 11. 
I had to make a very minor adjustment to one of the test.

aglkm:
I ran into the following issues.

### 1.

```
./grin-wallet info
Password: 
20250304 15:13:46.799 ERROR grin_wallet_impls::lifecycle::seed - wallet seed file /home/user/.grin/main/wallet_data/wallet.seed could not be opened (grin-wallet init). Run "grin-wallet init" to initialize a new wallet.
```
In my `grin-wallet.toml` file I have a different `data_file_dir` path but it doesn't take into account:
```
#where to find wallet files (seed, data, etc)
data_file_dir = "/home/user/grin/temp/wallet_data"
```

### 2.

`./grin-wallet -t d0/d1/d2 init` command creates 2 directories:

`ls -l`
```
drwxrwxr-x 3 user user     4096 Mar  4 15:30  d0
drwxrwxr-x 2 user user     4096 Mar  4 15:29 'd0\d1\d2'
```

### 3.

I'm not sure about changing to relative paths. What about API or 3rd party software relying on the `grin-wallet.toml` paths?

yeastplume:
Tests passing, looks good. @aglkm if you have time, can you see if those issues are still occurring in master and create a separate issue for them to keep them straight.

aglkm:
This PR was not ready for merging.
In it's current state it causes more issues than it fixes.
All pointed problems are still true. https://github.com/mimblewimble/grin-wallet/pull/731#issuecomment-2697447255

yeastplume:
> This PR was not ready for merging. In it's current state it causes more issues than it fixes. All pointed problems are still true. [#731 (comment)](https://github.com/mimblewimble/grin-wallet/pull/731#issuecomment-2697447255)

I'm usually okay with a bit of instability on master when it's just around the fringes of config, but in this case I think you're right and I was a bit too hasty merging. Going to revert and consider a few things here more carefully.

Anynomouss:
@aglkm 
#1
I see what you mean. Interestingly enough I did not change how the _data_path_ gets updated, so that is a mistake that must have been in the code already. I will fix it. I will also some test on Linux since what you describe in 2 does not happen on Windows.

#2
On Windows the paths is created with any number of sub-directories in that path. Perhaps on Linux this be different than on Windows. I have to check myself what you exactly mean with that two paths are created. Probably it is because there are multiple functions that create a path when it is not there and since Linux might work differently, e.g. paths instead of dirs. I already have my suspicions (its default.rs who is naughty).

#3
There is no switch to relative paths, nothing changed. Let me explain it since it is rather confusing:
_Original the wallet was updated by defaults.rs (made by @yeastplume), then @quentinlesceller implemented a better more explicit config creation with better error handling (and using relative paths) initiated by _wallet.rs_. Unfortunately the config did not get passed along properly and therefore was always discarded and overwritten by defaults.rs (which uses absolute paths). In the end I chose to stick with the original setup of absolute paths._

I see I need to do more fixing and testing, I will let you know when you can give it another test @aglkm

aglkm:
Issue `1` can only be re-created with your code changes, no such problem with the old code.

There's more to add to issue `2`. I see the following in `grin-wallet.toml` file:
`api_secret_path = "/home/user/grin/temp/d0\\d1\\d2/.owner_api_secret"`

I do see the changes related to absolute vs relative paths (issue `3`):
Before:
```
#where to find wallet files (seed, data, etc)
data_file_dir = "/home/user/grin/temp/top_dir_before/wallet_data"
```
After:
```
#where to find wallet files (seed, data, etc)
data_file_dir = "top_dir"
```

Anynomouss:
@aglkm I also noticed the replacement not working. I must have reverted something at some point or the change were only locally, I would have sworn it worked at some point. I do make the replacement in the code  but it does not work for some reason, probably because it contains escape characters.

In any case thank you for taking the time to do testing, it is appreciated. 
This time I will do more extensive testing myself on more than one machine before asking for a review.

Anynomouss:
I am going to take more time for this pull request. I have two versions with different dependency versions giving different results ( one with bug 1 fixed and 3 not working and the other exactly the other way) while having the same code. Also one of them cannot make push request to github anymore.
In addition the wallet generates some messages with the relative path later on while in the config we use absolute paths creating extra confusion. Rust is more capricious than I thought.

Issue #581, opened 2021-01-17
Deprecation of some API breaks mining on remote nodes 


slavikus:
**Describe the bug**
The recent [PR](https://github.com/mimblewimble/grin-wallet/pull/537) breaks the ability to mine on a node that runs on a remote machine.

`
ERROR grin_servers::mining::mine_block - Failed to get coinbase from http:/x.x.x.x:3415. Is the wallet listening? Request error: Cannot make request: error trying to connect: tcp connect error: Connection refused (os error 111)
ERROR grin_servers::mining::mine_block - Error building new block: Can't connect to wallet listener at "http://x.x.x.x:3415"; will retry
`

bladedoyle:
Known issue.
@jaspervdm  Please comment - what is the plan?

bladedoyle:
@slavikus As a workaround you can edit the source code and fix with something like this:
```sed -i 's/127.0.0.1:{}/0.0.0.0:{}/' config/src/types.rs```

jaspervdm:
We will likely have to introduce a new api type that solely exposes an endpoint to build a coinbase output. In the meanwhile it is recommended to use a reverse proxy in front of the wallet.

bladedoyle:
What happened is:
The wallets entire foreign_api was moved to tor listener and hardecoded to localhost

What should have happened:
A new receive_api added to the wallet, hardecoded to listen on localhost and used only by tor.   The receive_tx & finalize_tx methods should have been moved to this new api, and no longer available on the foreign_api. 
The foreign_api should have been left alone except for removing receive endpoints

PR #732, opened 2025-03-08
Ready for review: Fix for many wallet config errors including solution for -h and -t wrongly setting the node path

Anynomouss:
https://github.com/mimblewimble/grin-wallet/issues/478 [Partly Fixed/answered, cannot be solved easily since generate by toml crate ]
https://github.com/mimblewimble/grin-wallet/issues/728 [fixed]
https://github.com/mimblewimble/grin/issues/3803 [fixed]
https://github.com/mimblewimble/grin/issues/3394 [fixed]
https://github.com/mimblewimble/grin/pull/3420 [Abandoned Pull request, my pull replace this one]
https://github.com/mimblewimble/grin/issues/3002 [Partly fixed, not default values part]
https://github.com/mimblewimble/grin/issues/2300 [Could be closed, not dependent on my pull]

Note
The wallet firsts checks a) current directory for config, b) top_dir for config, then c) assumes its in home/.grin.
Option a) suppresses generation of a config file when running grin-wallet.exe init, not sure if that is intentional behavior or that the config should be crated but the one in current dir should be loaded.


Anynomouss:
**Update**
https://github.com/mimblewimble/grin-wallet/issues/581
https://github.com/mimblewimble/grin-wallet/issues/728 [fixed]
https://github.com/mimblewimble/grin/issues/3803 [fixed]
https://github.com/mimblewimble/grin/issues/3394 [fixed]
https://github.com/mimblewimble/grin/pull/3420 [Pull, my own pull replaces this one]
https://github.com/mimblewimble/grin/issues/3002 [Partly fixed, not default values part]
https://github.com/mimblewimble/grin/issues/2300 [Could be closed, not dependent on my pull]
•	Fix/workaround for a “bug” in the `fs::create_dir_alll` create to not recognize backslashes as directory separator on Linux. Fixed by converting the top_dir to always use OS_SEPARATOR. This bug only occurs on Linux and was uncovered by @aglkm
•	Fix all Seed generation message, to format the path with OS specific separator in seed.rs and default.rs
•	Removing config exists check from wallet_args, all logic is now handled by the cofig::initial_wallet_setup
•	Added smart functionality: If run with a config in the current directory, the wallet will load it and if run with ‘init’ will use as a template for config generation

**Possible things to improve**

1. I did not manage to change the top_dir arguments stored in the arguments, I have to update at multiple places in the code. 
Therefore these is a lot of formatting going on. Perhaps there are ways to reduce some of these formatting cals. I use a simple helper function to format the path to use OS_SEPARATOR. This helper function is used in seed.rs, defaults.rs and wallet.rs, each with a new definition of the function. If there is such things as a crate where other helper functions are put, perhaps I could avoid redefining it.

2. Config generation is now handled by wallet.rs. This means that if defaults.rs is called in another way to generate a config, e.g. owner API, the behaviors is as before not distinguishing node and wallet directory. It needed I can also adjust the code in default.rs to use `config::get_node_path` and `config::get_wallet_path` , but I would have to convert the Config Errors to normal Errors. It here are reasons to make change there as well I could still do it, but if not needed I will move on to look at the node header syncing.

@aglkm If you have time, can you give it another test?

aglkm:
Issues reported here: https://github.com/mimblewimble/grin-wallet/pull/731#issuecomment-2697447255
Fixed: 2 and 3.
Not fixed: 1.

Steps to re-create:
1. Compile the wallet with the PR
2. Init a wallet: ./target/release/grin-wallet -t top-dir init
3. Copy executable into the top dir: cp target/release/grin-wallet ./top-dir/.
4. CD into the top dir: cd top-dir/
5. Make sure there's no wallet seed in the default directory (/home/user/.grin/main/wallet_data/) and run the wallet: ./grin-wallet info
You will see the following:
```
20250518 12:32:59.411 ERROR grin_wallet_impls::lifecycle::seed - wallet seed file /home/user/.grin/main/wallet_data/wallet.seed could not be opened (grin-wallet init). Run "grin-wallet init" to initialize a new wallet.
```

So the wallet is ignoring the "top-dir" directory (/home/user/grin/grin-wallet-sources/grin-wallet-pr732/grin-wallet/top-dir) and still is trying to use the default one.

In grin-wallet.toml file the path is correct:
```
#where to find wallet files (seed, data, etc)
data_file_dir = "/home/user/grin/grin-wallet-sources/grin-wallet-pr732/grin-wallet/top-dir/wallet_data"
```

Another issue:
Make sure you are one level up from the "top-dir" directory, then run: ./top-dir/grin-wallet -t top-dir info
You will get:
```
Unable to load wallet configuration: Error serializing configuration: /home/user/grin/grin-wallet-sources/grin-wallet-pr732/grin-wallet/top-dir/grin-wallet.toml already exists in the target directory (/home/user/grin/grin-wallet-sources/grin-wallet-pr732/grin-wallet/top-dir/wallet_data). Please remove it first (Run `grin-wallet init` to create a new wallet)
```

Anynomouss:
@aglkm Thanks for explaining issue 1, I misunderstood you before. 

Basically grin-wallet will always assume that if you do not provide the  '-t' topi-dir command that it should search for those files in the default directory. What you are basically proposing is a **auto-detect wallet_dat function** to first test if the current directory contains a `wallet_data` folder, if not default to the users home directory. This would be new wallet behavior, I could implement it. What do you think @yeastplume, would that be desired behavior?
The only downside I can think of is if you would run `grin-wallet init -h`, this means automatically that if you would thereafter run grin-wallet info the wallet would be so smart to automatically use this new created wallet and not your home directory. This is smart behavior, but if I would implement I think I should give a warning to the users. Something along the line `'wallet_data' detected in current dir, defaulting to use this wallet`. 
What do you think @yeastplume should this smart behavior be added, or just keep things explicit?
```
Another issue:
Make sure you are one level up from the "top-dir" directory, then run: ./top-dir/grin-wallet -t top-dir info
You will get:
```
This is actually desired behavior. What I mean with it is that all paths you provide should be given relative to your current working directory, not relatively to the executable. Additionally this allows power users to do some tricks. For example:
1) I run `grin-wallet.exe  -t top_dir init` (with a template config file in the current directory of _grin-wallet.exe_
_What will happen is that grin-wallet will create a new config file in the top-dir which uses the config file in the working directory as a template, for example keeping specific peers, ports etc. in the config file. Only the node path, wallet path and tor path will be updated._
2) cd top_dir
3) run  ..grin-wallet.exe  -t ../top_dir info
_Now there will be no template file in your current working directory since you are in directory `top-dir`, so the wallet will run using your new config file_

I could change this behavior, but I think it is customary for executable to use path relative to the current working directory and relative to the directory the executable is in. What do you think @yeastplume?

aglkm:
It is documented in grin-wallet docs that the wallet will check the current directory first and then the
default one.
```
When running any grin-wallet command, grin will check the working directory if these files exist. If not, it will use the default location ~/.grin.
```

This is how it works currently.

What is this PR proposing: if you wish to use local (non default) config files, you have to explicitly specifying current working directory every time you execute a wallet, while you are already locating there. I disagree with this.

```
./top-dir/grin-wallet -t top-dir info
```

Here ☝️ I've specified the path to grin-wallet executable relatively to the current working directory and provided the path to "top-dir" relatively to the current working directory as well, not an executable. The current wallet behaviour has no such issue.

You wrote:
```
What will happen is that grin-wallet will create a new config file in the top-dir which uses the config file in the working directory as a template, for example keeping specific peers, ports etc. in the config file. Only the node path, wallet path and tor path will be updated.
```

Wallet shouldn't try to think instead of a user, because a desired behaviour could be different. I could want a fresh config file, but instead wallet will create it from an old "template" - a config file which is placed in the current directory. It's more straightforward to let user to copy the old config file if he wishes to preserve the settings.

You wrote:
```
Now there will be no template file in your current working directory since you are in directory top-dir, so the wallet will run using your new config file
```

The difference between a template file and a config file: a config file becoming a template file if the user is trying to create a new wallet while also the config file exists in the current directory. If we want to have "template" files it could be better to introduce a new option, so the user can directly specify what he wants to do and do not let the wallet to decide.

As a user, it could be confusing, that the wallet is silently trying to use a config file as a template file.

There's another thing that need attention. When you are creating a config file from a "template" it will create a relative path in grin-wallet.toml file, like that:
```
#where to find wallet files (seed, data, etc)
data_file_dir = "top_dr"
```

What's the purpose of this change? This could potentially break the things, e.g. API.

Anynomouss:
> It is documented in grin-wallet docs that the wallet will check the current directory first and then the
> default one.

Interesting, this functionality is not present in the code. It does make sense to add it since many users might expect the wallet to be smart enough to detect it. I will add it to the code, it is similar smart behaviour as I have implemented for detecting the path to the node.

> Wallet shouldn't try to think instead of a user, because a desired behavior could be different. I could want a fresh config file, but instead wallet will create it from an old "template" - a config file which is placed in the current directory. It's more straightforward to let user to copy the old config file if he wishes to preserve the settings. 

I disagree, what you are saying now is that you do not want smart behavior, while you do want smart behavior for the wallet to detect the path to the node and detect if there is a local wallet folder, that is contradicing. The whole point of this pull request is to fix `-t `and `-here` and adding smart behavior to grin-wallet to let it behave like the user expect it should:

- Looking where to find the node, first in the current directory, if not there in the default directory
- The wallet should look in the current directory first to see if the wallet is there, this is the "smart behavior" <should have been there already according to the documentation>
- Detecting any custom config file in the executable path and using its settings when opening the wallet

Using a config in the executable dir with -init as template is an extension of the behavior  that when running without -init the config file will indicate which wallet to open and with what settings. Again, this is not me making real change, just extending on existing behavior.

> As a user, it could be confusing, that the wallet is silently trying to use a config file as a template file.
I think this is the case for all smart behavior. All smart behavior is implicit since the user does not explicitly have to state it. The solution is simple, I will ad _Warning_ or _Info_ level messages to the user to notify them this smart behavior is triggered.
The alternative is not having any smart behavior at all which is very, very bad for the user experience and conflicts with the behavior described in the readme file. Also we have to be realistic, what is the chance of a user copying a config file in the  working directory of without knowing about this smart behavior? I think it is very small, add a message to inform the user about this smart behavior and we should be fine. More importantly, lets make grin-wallet user friendly for those 99% of user that do expect the _grin-wallet_ to have emulated some minimal intelligence.

Anynomouss:
Config fix
Latest pull

https://github.com/mimblewimble/grin-wallet/pull/732

old pull

https://github.com/mimblewimble/grin-wallet/pull/731

Files changed
config/src/config.rs
controller/src/command.rs
impls/src/lifecycle/default.rs
impls/src/lifecycle/seed.rs
src/bin/grin-wallet.rs
impls/src/node_clients/http.rs

I think the pull is ready for review, however for whatever reason git or Github shows all files as changed even though the content has not changed at all. I suspect it change a line endings.

**Files that have changed (others can be discarded)**
config/src/config.rs
impls/src/lifecycle/default.rs
impls/src/lifecycle/seed.rs
src/bin/grin-wallet.rs
impls/src/node_clients/http.rs

The above files integrate the proposed changes in https://github.com/mimblewimble/grin-wallet/pull/732 for the above mentioned files

**What this pull request fixes:**

- The main changes of this pull request are that running a) `grin-wallet.exe -t top-dir init` b) `grin-wallet.exe init -h` c) grin-wallet.exe -t top-dir info, will use (and if needed create) the proper directory.
- Input paths provided as input using -t (top-dir) will be a) made absolute b) any  backslash "\\" is replaced with a "/" slash to avoid errors on Linux, Prefix added by Windows 10/11 ("\\?\") is removed to maintain platform compatibility of grin-wallet.toml
In short both using relative or absolute paths as input for -t will work now.
- The wallet path is selected as follows a) -t top directory if argument provided, b) config in working dir c) home
- Grin-wallet prints the wallet directory it is using to the console. Although this was mostly intended for debugging, it might be useful behavior, even if just to check if this pull request works as intended.
 - Previously this pull request introduced using a `grin-wallet.toml` in the current directory as template. After further consideration I removed this functionality since it might cause confusion as @aglkm pointed out and it does not have any clear use case.

aglkm (commented):
Please fix this compilation warning:

warning: function `check_config_current_dir` is never used
   --> config/src/config.rs:123:4
    |
123 | fn check_config_current_dir(path: &str) -> Option<PathBuf> {
    |    ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

config/src/config.rs
```diff
@@ -14,60 +14,112 @@
...
+pub const GRIN_HOME: &str = ".grin";
 /// Wallet data directory
 pub const GRIN_WALLET_DIR: &str = "wallet_data";
 /// Node API secret
 pub const API_SECRET_FILE_NAME: &str = ".foreign_api_secret";
 /// Owner API secret
 pub const OWNER_API_SECRET_FILE_NAME: &str = ".owner_api_secret";
 
-fn get_grin_path(
+/// Function to locate the wallet dir and wallet.toml in the order
```
aglkm, line 45: The file is called grin-wallet.toml

config/src/config.rs
```diff
@@ -14,60 +14,112 @@
...
 }
 
+/// Smart function to detect the the nodes .api_secret in the order
+/// a) top-dir, b) home directory, create directory if needed
+pub fn get_node_path(
+	data_path: Option<PathBuf>,
+	chain_type: &global::ChainTypes,
+) -> Result<PathBuf, ConfigError> {
+	let node_path = match data_path {
+		// 1) A If top dir provided and api_secret exist, return top dir
```
aglkm, line 88: Same. It could be unclear what's this api_secret (owner, foreign or both?). Same in other places.

I see we call it API_SECRET_FILE_NAME in the code, so it's already a bit confusing. Anyways, adding more context could be useful.

DavidBurkett:
It looks like there's a lot of history here. Can you summarize your changes so I know which behavior changes are intentional, and which may be unintentional? For example, is it intentional that `grin-wallet init` would no longer honor an existing `./grin-wallet.toml`?

Anynomouss:
@DavidBurkett  I described the changes in this post:
https://github.com/mimblewimble/grin-wallet/pull/732#issuecomment-3307000075

TLDR; The PR fixes a problem with creating a new wallet using a custom directory using either -h or -t e.g. ` ./grin-wallet.exe  init -h` or  `./grin-wallet.exe  init -t dir1/dir2`,  which did not update the path in  `./grin-wallet.toml.` In addition it improves file path handling, and printing of file paths in some-errors. The code does not change how config files are generated, migrate, loaded etc. and should not affect compatibility with older configs in any way. 

There is quite some history with regards to `grin-wallet.toml` generation which we can discuss in Telegram if needed. Originally the config was generated later in the wallet creation process with a rather long call chain (I think Yeast set it up this way), Quentin changed it to be done early in the process its handled by `main.rs`. He did not however properly set the path in the config since there was no separation between node and wallet paths. This PR aims to fix that. I merged the upstream commits that fixes the failed Windows integration test.

impls/src/lifecycle/default.rs
```diff
@@ -178,11 +198,11 @@ where
 	) -> Result<(), Error> {
 		let mut data_dir_name = PathBuf::from(self.data_dir.clone());
 		data_dir_name.push(GRIN_WALLET_DIR);
-		let data_dir_name = data_dir_name.to_str().unwrap();
+		let data_dir_name = fmt_path((data_dir_name.to_str().unwrap()).to_string());
 		let exists = WalletSeed::seed_file_exists(&data_dir_name);
 		if !test_mode {
 			if let Ok(true) = exists {
-				let msg = format!("Wallet seed already exists at: {}", data_dir_name);
+				let msg = format!("Wallet seed already exists at4565: {}", data_dir_name);
```
DavidBurkett: Remove the 4565. I'm guessing that was for debugging.

Anynomouss: I do not remember adding that, but I guess it was for testing, or I accidentality typed a 2FA code in the wrong screen 😅. In any case, it should be removed.

impls/src/lifecycle/seed.rs
```diff
@@ -166,10 +182,15 @@ impl WalletSeed {
 		// create directory if it doesn't exist
 		fs::create_dir_all(data_file_dir).map_err(|_| Error::IO)?;
 
-		let seed_file_path = &format!("{}{}{}", data_file_dir, MAIN_SEPARATOR, SEED_FILE,);
-
+		let seed_file_path = &format!(
```
DavidBurkett, line 185: nit: You normalize _after_ calling `fs::create_dir_all`, so you could create a different directory than the one you try to write to.  It's probably not a likely issue, but we should come back and clean this up later if we get a chance.

Anynomouss: Actually, would it be possible to normalize the path directly from the input parameter for the directory? It would be best and could avoid a lot of redundant code, I just did not know how to get it done in such a way.

DavidBurkett: Let's just get this merged for now. Address the other comment and then you can come back and do this in another PR, if you don't mind?

PR #738, opened 2025-05-31
Provide proxy on node client creation

ardocrat:
With current implementation we could not provide proxy configuration to `HTTPNodeClient` on wallet creation for RPC communication, also there was hardcoded `socks5h` scheme used for Tor client at another place.


yeastplume (approved):
LGTM, thanks!

PR #748, opened 2026-04-10
Save last scanned block info for wallet scanning

ardocrat:
- Save last `ScannedBlockInfo` on collect outputs every 10k blocks on scan.

### DIFF
--- a/libwallet/src/api_impl/owner.rs
+++ b/libwallet/src/api_impl/owner.rs
@@ -1111,21 +1111,13 @@ where
 	let last_scanned_block = {
 		wallet_lock!(wallet_inst, w);
 		match w.init_status()? {
-			WalletInitStatus::InitNeedsScanning => {
-				w.last_scanned_block().unwrap_or(ScannedBlockInfo {
-					height: 0,
-					hash: "".to_owned(),
-					start_pmmr_index: 0,
-					last_pmmr_index: 0,
-				})
-			}
 			WalletInitStatus::InitNoScanning => ScannedBlockInfo {
 				height: tip.clone().0,
 				hash: tip.clone().1,
 				start_pmmr_index: 0,
 				last_pmmr_index: 0,
 			},
-			WalletInitStatus::InitComplete => w.last_scanned_block()?,
+			_ => w.last_scanned_block()?,
 		}
 	};
 
--- a/libwallet/src/api_impl/owner.rs
+++ b/libwallet/src/api_impl/owner.rs
@@ -983,6 +983,7 @@ where
 		wallet_inst.clone(),
 		keychain_mask,
 		delete_unconfirmed,
+		None,
 		start_height,
 		tip.0,
 		status_send_channel,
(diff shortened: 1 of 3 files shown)


ardocrat:
For some reason CI failed at not commit-related step, same step passed at fork's commit: https://github.com/GetGrin/wallet/actions/runs/24289586705/job/70924613337

libwallet/src/internal/scan.rs
```diff
@@ -264,9 +275,24 @@ where
...
 			status_send_channel,
-			perc_complete as u8,
+			perc_complete,
 		)?);
 
+		// Save last scanned block info to continue from same index in case of interruption.
+		{
+			wallet_lock!(wallet_inst, w);
+			let mut batch = w.batch(*keychain_mask)?;
+			batch.save_last_scanned_block(ScannedBlockInfo {
```
DavidBurkett: Critical: You're saving the checkpoint before we've applied the outputs to the wallet state. If the process is interrupted during this batch of outputs, restarting will end up skipping this batch entirely.

ardocrat: indeed, will make scanning in batches to save some parts time to time.

libwallet/src/api_impl/owner.rs
```diff
@@ -974,25 +975,30 @@ where
...
-		keychain_mask,
-		delete_unconfirmed,
-		start_height,
-		tip.0,
-		status_send_channel,
-	)?;
-	info.hash = tip.1;
+	// Scan every 10k heights to save data between batches in case of interruption.
+	let mut total_pmmr_range = None;
+	for h in (start_height..tip.0).step_by(10001) {
```
wiesche89, line 982: this is an edge case we should fix before merging!? The current range excludes tip.0, so the last tip block can be skipped at batch boundaries.

example:
for h in (start_height..=tip.0).step_by(10001) {

libwallet/src/api_impl/owner.rs
```diff
@@ -1135,18 +1140,23 @@ where
...
-		false,
-		start_index,
-		tip.0,
-		status_send_channel,
-	)?;
-
-	info.hash = tip.1;
+	// Scan every 10k heights to save data between batches in case of interruption.
+	let mut total_pmmr_range = None;
+	for h in (start_height..tip.0).step_by(10001) {
```
wiesche89, line 1155: same context (line 982)

example:
for h in (start_height..=tip.0).step_by(10001) {

libwallet/src/api_impl/owner.rs
```diff
@@ -974,25 +975,30 @@ where
...
+			keychain_mask,
+			delete_unconfirmed,
+			h,
+			cmp::min(tip.0, h + 10000),
+			start_height,
+			tip.0,
+			total_pmmr_range,
+			status_send_channel,
+		)?;
+		info.hash = tip.1.clone();
```
wiesche89: Could this be inconsistent for intermediate batches? The saved height is the batch end, but the hash is always from the final tip.

libwallet/src/api_impl/owner.rs
```diff
@@ -1110,23 +1116,22 @@ where
...
 				start_pmmr_index: 0,
 				last_pmmr_index: 0,
 			},
-			WalletInitStatus::InitComplete => w.last_scanned_block()?,
+			_ => w.last_scanned_block()?,
 		}
 	};
 
-	let start_index = last_scanned_block.height.saturating_sub(100);
+	let start_height = last_scanned_block.height.saturating_sub(100);
```
wiesche89: @Anynomouss had a good point!
Could this be a problem after a reorg and restart? With batch checkpoints, the next scan only goes back 100 blocks. Should this be a larger safety window, for example 48 hours worth of blocks (24 * 60 * 2= 2880)?

example:
const REORG_RESCAN_WINDOW: u64 = 24 * 60 * 2;

ardocrat: interesting point, this can only increase time to sync, will test for how long

PR #750, opened 2026-05-22
Integrated Tor arti Client

ardocrat:
- Integrated Arti client for listener (https://gitlab.torproject.org/tpo/core/arti/)
- Bridges support
- Added `bridge_bin_path` and `use_integrated` options at config
- Update `tokio` to last version
- Use Arti client to send requests
- Do not try to send if `skip_send_attempt` presents at config


impls/src/tor/arti.rs
```diff
@@ -0,0 +1,311 @@
...
+		.map_err(|e| Error::TorProcess(format!("{:?}", e)))?;
+	let c = client.clone();
+	let r = runtime.clone();
+	let res = thread::spawn(move || {
+		r.block_on(async move {
+			let res = c.bootstrap().await;
+			if res.is_ok() {
+				let mut percent = 0.0;
+				let mut prev_percent = 0.0;
+				while percent != 1.0 {
```
wiesche89: Comparing a floating point value with != 1.0 is unsafe, as it may never reach exactly 1.0

fix in: while percent < 1.0 {

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,311 @@
...
+					.protocols(vec![t
+						.parse()
+						.map_err(|e| Error::TorConfig(format!("{:?}", e)))?])
+					.path(CfgPath::new(bin_path))
+					.run_on_startup(true);
+				builder.bridges().transports().push(transport);
+			}
+		}
+	}
+	let config = builder.build().unwrap();
```
wiesche89: We should return a Tor-related error here

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,311 @@
...
+/// Save Onion service key to keystore.
+fn add_service_key(
+	mistrust: &Mistrust,
+	key: &SecretKey,
+	hs_nickname: &HsNickname,
+	path: PathBuf,
+) -> tor_keymgr::Result<()> {
+	let arti_store = ArtiNativeKeystore::from_path_and_mistrust(path, mistrust)?;
+
+	let key_manager = KeyMgrBuilder::default()
```
wiesche89, line 365: Same issue, panics if the keystore directory is not writable.

controller/src/controller.rs
```diff
@@ -343,15 +346,14 @@ where
...
-		};
-		warn!("Slatepack Address is: {}\n{}", a, qr_string);
-	}
+
+	let sp_address = SlatepackAddress::try_from(onion_address.clone())?;
+	let qr_string = match QrCode::new(sp_address.to_string()) {
+		Ok(qr) => qr.to_string(false, 3),
+		Err(_) => "Failed to generate QR code!".to_string(),
+	};
+	warn!("Slatepack Address is: {}\n{}", sp_address, qr_string);
```
wiesche89: If Tor fails to start, the wallet still shows the onion address to the user. They might share it with someone expecting to receive funds, but nothing will arrive. Please only show the address when Tor actually started successfully.  Or add a clear warning like "Tor failed to start , this address is currently unreachable"?

ardocrat: Error is already showing, will not show slatepack address in this case, will show http listener address after api launch.

impls/src/tor/arti.rs
ardocrat: Added handling of keystore errors

controller/src/controller.rs
```diff
@@ -343,14 +345,17 @@ where
 	let api_thread = apis
 		.start(socket_addr, router, tls_config, api_chan)
 		.map_err(|_| Error::GenericError("API thread failed to start".to_string()))?;
-
 	warn!("HTTP Foreign listener started.");
-	if let Some(a) = address {
-		let qr_string = match QrCode::new(a.to_string()) {
+
+	if let Some(_) = tor_service.ok() {
```
wiesche89: I think this drops the Tor service too early. The Slatepack address can be printed, but the onion service may already be stopped before the listener keeps running. It also looks like this can print the address even when Tor is disabled. We should keep the Tor handle alive for the whole listener lifetime and only show the address when Tor actually started.

ardocrat: Fixed

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,317 @@
...
+		.map_err(|e| Error::TorProcess(format!("{:?}", e)))?;
+	let c = client.clone();
+	let r = runtime.clone();
+	let res = thread::spawn(move || {
+		r.block_on(async move {
+			let res = c.bootstrap().await;
+			if res.is_ok() {
+				let mut percent = 0.0;
+				let mut prev_percent = 0.0;
+				while percent < 1.0 {
```
wiesche89: Without a timeout, this loop can hang the wallet forever if tor has bootstrapped but the status never reaches 100%. Timeout wait timer 60sec?

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,317 @@
...
+				.await
+				.map_err(|e| Error::TorProcess(format!("{:?}", e)))?;
+			let body = body_resp.to_bytes().into();
+			let body_text =
+				String::from_utf8(body).map_err(|e| Error::TorProcess(format!("{:?}", e)))?;
+			Ok(body_text)
+		})
+	})
+	.join()
+	.unwrap();
```
wiesche89: If the spawned thread panics for any reason, unwrap will crash the wallet process. return a error?

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,317 @@
...
+	let (client, _) = init_client(&state_path, &cache_path, config)?;
+	Ok(Tor {
+		process: None,
+		service: None,
+		client: Some(client),
+	})
+}
+
+/// Make POST request with provided client.
+pub fn tor_post<IN>(
```
wiesche89, line 159: There's no timeout here in this function, if the recipient doesn't respond, the send will hang forever instead of returning an error. Please add a timeout.

impls/src/tor/arti.rs
ardocrat: added bootstrap timeout

impls/src/tor/arti.rs
ardocrat: returning error at this case

impls/src/tor/arti.rs
ardocrat: added 60s timeout

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,339 @@
...
+	let url = url.to_string();
+	let url: Uri = url
+		.parse()
+		.map_err(|_| Error::GenericError("Bad URL".to_owned()))?;
+	let res: Result<String, Error> = thread::spawn(move || {
+		let c = client.clone();
+		client.runtime().block_on(async move {
+			let res = c
+				.runtime()
+				.timeout(Duration::from_secs(REQUEST_TIMEOUT_MS), async {
```
wiesche89: This should use from_millis here :)

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,339 @@
...
+	Ok(Tor {
+		process: None,
+		service: Some(running_onion_service),
+		client: Some(client),
+	})
+}
+
+/// Start Tor client to send requests.
+pub fn start_tor_client(tor_dir: &str, config: TorConfig) -> Result<Tor, Error> {
+	let state_path = Path::new(tor_dir).join("state");
```
wiesche89, line 144: Do we need to delete cache and state on every normal start, or would it be enough to do a clean retry only after bootstrap fails or times out?

With my address checker the bootstrap already takes noticeably longer because of this. But I am also fine with it if we want to be on the safe side.

impls/src/tor/arti.rs
ardocrat: indeed, forgot change after test

impls/src/tor/arti.rs
ardocrat: Now we store arti runtime to not panic after instance drop, cache cleaning is not needed anymore

impls/src/tor/mod.rs
```diff
@@ -12,7 +12,31 @@
...
+
+/// Arti client runtime wrapper.
+pub struct ArtiRuntimeWrapper {
+	pub runtime: tor_rtcompat::tokio::TokioNativeTlsRuntime,
+}
+
+impl Default for ArtiRuntimeWrapper {
+	fn default() -> Self {
+		Self {
+			runtime: tor_rtcompat::tokio::TokioNativeTlsRuntime::create().unwrap(),
```
wiesche89: this unwrap could still panic if the runtime cannot be created. Can we return a normal Tor error here instead?

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,345 @@
...
+						.await
+						.map_err(|e| Error::TorProcess(format!("{:?}", e)))?;
+					let (mut request_sender, connection) =
+						hyper::client::conn::http1::handshake(TokioIo::new(stream))
+							.await
+							.map_err(|e| Error::TorProcess(format!("{:?}", e)))?;
+
+					// Spawn a task to poll the connection and drive the HTTP state.
+					tokio::spawn(async move {
+						connection.await.unwrap();
```
wiesche89: Could we avoid the unwrap here? If the connection ends with an error, this would panic inside a Tokio worker. Logging the error would be safer.

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,345 @@
...
+	// Launch client.
+	let r = ARTI_RUNTIME.runtime.clone();
+	let client = TorClient::with_runtime(r)
+		.config(config.clone())
+		.create_unbootstrapped()
+		.map_err(|e| Error::TorProcess(format!("{:?}", e)))?;
+	let c = client.clone();
+	let res = client.runtime().block_on(async move {
+		let bootstrap = async || {
+			if c.bootstrap().await.is_ok() {
```
wiesche89: If bootstrap() fails here, we still continue as if Tor started successfully. That could leave us with an unbootstrapped client.

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,345 @@
...
+	config: TorConfig,
+) -> Result<Tor, Error> {
+	let state_path = Path::new(&tor_dir).join("state");
+	let cache_path = Path::new(&tor_dir).join("cache");
+	let (client, config) = init_client(&state_path, &cache_path, config)?;
+
+	// Add service key to keystore.
+	let onion_address =
+		OnionV3Address::from_private(&key.0).map_err(|e| Error::TorConfig(format!("{:?}", e)))?;
+	let hs = HsNickname::new(onion_address.to_string()).unwrap();
```
wiesche89: Could we add proper error handling here instead of unwrapping?

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,345 @@
...
+	let keystore_path = Path::new(&state_path).join("keystore");
+	let _ = add_service_key(config.fs_mistrust(), &key, &hs, keystore_path)?;
+
+	// Launch Onion service.
+	let service_config = OnionServiceConfigBuilder::default()
+		.nickname(hs.clone())
+		.build()
+		.unwrap();
```
wiesche89: Could we add proper error handling here instead of unwrapping?

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,345 @@
...
+	nickname: HsNickname,
+) -> Result<(), Error>
+where
+	S: futures::Stream<Item = tor_hsservice::RendRequest> + Unpin + Send + 'static,
+{
+	let runtime = client.runtime().clone();
+
+	// Setup proxy to forward request from Tor address to local address.
+	let proxy_rule = ProxyRule::new(
+		ProxyPattern::one_port(80).unwrap(),
```
wiesche89: Could we add proper error handling here instead of unwrapping?

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,345 @@
...
+		ProxyAction::Forward(Encapsulation::Simple, TargetAddr::Inet(addr)),
+	);
+	let mut proxy_cfg_builder = ProxyConfigBuilder::default();
+	proxy_cfg_builder.set_proxy_ports(vec![proxy_rule]);
+	let proxy = OnionServiceReverseProxy::new(proxy_cfg_builder.build().unwrap());
```
wiesche89: Could we add proper error handling here instead of unwrapping?

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,345 @@
...
+		.build()
+		.map_err(|e| Error::TorProcess(format!("{}", e)))?;
+
+	let expanded_sk =
+		ExpandedSecretKey::from_bytes(Sha512::default().chain_update(key).finalize().as_ref());
+
+	let mut sk_bytes = [0_u8; 64];
+	sk_bytes[0..32].copy_from_slice(&expanded_sk.scalar.to_bytes());
+	sk_bytes[32..64].copy_from_slice(&expanded_sk.hash_prefix);
+	let expanded_kp = ExpandedKeypair::from_secret_key_bytes(sk_bytes).unwrap();
```
wiesche89: Could we add proper error handling here instead of unwrapping?

impls/src/tor/mod.rs
ardocrat: added error handling

impls/src/tor/arti.rs
ardocrat: showing an error at log here, if something will go wrong request should not pass

impls/src/tor/arti.rs
ardocrat: returning error at future now

impls/src/tor/arti.rs
ardocrat: added error handling

impls/src/tor/arti.rs
ardocrat: added error handling

impls/src/tor/arti.rs
ardocrat: added error handling

impls/src/tor/arti.rs
ardocrat: added error handling

impls/src/tor/arti.rs
ardocrat: added check for key emptiness

impls/src/adapters/tor.rs
wiesche89: This existed before, but this PR now uses it for the new default Arti path too. A bad remote response can still panic here, it would be better to return a clean error.

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,366 @@
...
+	}
+	let config = builder
+		.build()
+		.map_err(|e| Error::TorConfig(format!("{:?}", e)))?;
+
+	// Launch client.
+	let r = ARTI_RUNTIME
+		.as_ref()
+		.clone()
+		.unwrap_or(ArtiRuntimeWrapper::create()?)
```
wiesche89: The fallback runs every time, even when a runtime already exists. This creates unnecessary runtimes, please make it lazy.

controller/src/controller.rs
```diff
@@ -285,49 +269,68 @@ pub fn foreign_listener<L, C, K>(
...
-	let (_tor_process, address) = match use_tor {
-		true => {
-			match init_tor_listener(
-				wallet.clone(),
-				keychain_mask.clone(),
+	// Need to keep external process in scope while the listener is running.
+	let tor_service = if use_tor {
+		let use_integrated = tor_config.use_integrated.unwrap_or(true);
+		let res = if use_integrated {
+			start_tor_service(sec_key, &tor_dir, addr, tor_config.clone())
```
wiesche89, line 324: The HTTP listener waits for Tor here. With poor Tor connectivity startup can hang for up to 60s, HTTP could probably start first.

impls/src/adapters/tor.rs
```diff
@@ -12,136 +12,119 @@
...
 		} else {
-			Ok(HttpSlateSender {
+			let tor_dir = {
+				let mut path = PathBuf::from(&config.send_config_dir);
+				path.push("tor");
+				path.push("sender");
+				path
+			};
+			let tor = if config.use_integrated.unwrap_or(true) {
+				start_tor_client(tor_dir.to_str().unwrap(), config.clone())?
```
wiesche89, line 54: The Arti sender seems to bootstrap a fresh client for each send workflow. In long-running or batch use this can add repeated delays of up to 60s. Could we reuse a bootstrapped client?

ardocrat: we can keep tor client as global state, currently at CLI there is no such case, there is on request per command

impls/src/tor/arti.rs
ardocrat: we can try to record at state again and return an error otherwise

controller/src/controller.rs
```diff
@@ -343,14 +313,52 @@ where
...
+				error!("Tor Error: {}", e);
+				warn!("Listener is available on {}", addr);
+				Err(e)
+			}
+		}
+	} else {
+		Ok(None)
+	};
+
+	if tor_service.is_ok() {
```
wiesche89, line 353: This works as it stands, but the Tor service lifetime is a bit easy to break in a future refactor. Could we make the long-lived handle more explicit, or add a small regression test to cover it?

ardocrat: We can make global arti client state, external Tor process can not be stored globally

impls/src/adapters/tor.rs
ardocrat: added handling errors of json parsing

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,372 @@
...
+						}
+						prev_percent = percent;
+						tokio::time::sleep(Duration::from_millis(1000)).await;
+					}
+					Ok(())
+				}
+				Err(e) => Err(e),
+			};
+		};
+		match c
```
wiesche89, line 301: I think we should propagate the inner bootstrap error here. timeout() returns Ok(Err(e)) when bootstrap itself fails, and that currently gets treated as success. That can leave us with a Tor client that is reported as bootstrapped but is not actually usable.

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,372 @@
...
+		.parse()
+		.map_err(|_| Error::GenericError("Bad URL".to_owned()))?;
+	let res: Result<String, Error> = thread::spawn(move || {
+		let c = client.clone();
+		client.runtime().block_on(async move {
+			let res = c
+				.runtime()
+				.timeout(Duration::from_millis(REQUEST_TIMEOUT_MS), async {
+					let stream = c
+						.connect((url.host().unwrap(), url.port_u16().unwrap_or(80)))
```
wiesche89: After parsing the URL we still unwrap the host. It would be safer to return a normal wallet error here, so malformed URLs cannot panic the sender path.

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,372 @@
...
+		.map_err(|e| Error::TorConfig(format!("{:?}", e)))?;
+	let running_onion_service = match client.launch_onion_service(service_config) {
+		Ok(res) => {
+			if let Some((service, request)) = res {
+				// Launch service proxy.
+				let addr: SocketAddr = addr
+					.parse()
+					.map_err(|e| Error::TorProcess(format!("{:?}", e)))?;
+				let c = client.clone();
+				thread::spawn(move || {
```
wiesche89, line 104: If handle_requests returns while the wallet is still running, this proxy thread exits and the listener may still look healthy. Do we expect this to live for the whole service lifetime, or should we add a warning/restart path here?

impls/src/adapters/tor.rs
wiesche89: Agreed, for one shot CLI commands this is less visible. I was mainly thinking about Owner API, GUI or long running wallet usage, where multiple sends can happen in the same process.

impls/src/tor/arti.rs
ardocrat: added restart after 1 sec + error message show

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,396 @@
...
+/// Create Tor client.
+fn init_client(
+	state_path: &PathBuf,
+	cache_path: &PathBuf,
+	config: TorConfig,
+) -> Result<(Arc<TorClient<TokioNativeTlsRuntime>>, TorClientConfig), Error> {
+	let mut builder = TorClientConfigBuilder::from_directories(&state_path, cache_path);
+	builder.address_filter().allow_onion_addrs(true);
+
+	// Configure bridge.
```
wiesche89, line 244: For existing configs without the new use_integrated field, unwrap_or(true) makes us switch to the integrated Arti path automatically. I guess that’s intended, but it would be helpful to log clearly on startup whether we’re using integrated Arti or the external Tor process. That would make post-upgrade debugging much easier.

impls/src/tor/arti.rs
wiesche89: Thanks, this is better. One small thing: this only restarts on Err. If handle_requests returns Ok, the proxy task still exits. Could we handle both cases, maybe with a simple loop instead of the recursive Box::pin call?

ardocrat: If `handle_requests` returns Ok then thread will be exited (means proxy stopped at right way), it blocked before Ok, on error we just restart same function

impls/src/tor/arti.rs
```diff
@@ -0,0 +1,396 @@
...
+		match bridge.get_transport() {
+			None => {
+				return Err(Error::TorConfig(format!(
+					"No transport found at {}",
+					bridge_line
+				)))
+			}
+			Some(t) => {
+				// Now configure bridge transport. (Requires the "pt-client" feature)
+				let bin_path = config.bridge.bridge_bin_path.unwrap_or(t.to_owned());
```
wiesche89, line 259: This maps bridge_line and bridge_bin_path, but existing tor.proxy settings and bridge.client_option seem to be ignored in the Arti path. Could we either support them, fall back to external Tor, or return a clear unsupported-config error when they are set?

ardocrat: `client_option` is not needed for arti (from comments it used for snowflake) currently snowflake uses only one line setup, tor proxy allows to send requests over proxy, not related to arti

controller/src/controller.rs
wiesche89: Thanks, I re checked the current code and this looks fine now.

impls/src/tor/arti.rs
wiesche89: Makes sense for client_option, thanks. I’m still a bit unsure about tor.proxy though, since the external tor path uses it for tor’s upstream proxy reachable addresses. If Arti doesn’t support that yet, could we at least warn or error when tor.proxy is set with use_integrated=true?

ardocrat: currently we ignore proxy when integrated is enabled

wiesche89: Got it. In that case, could we add a warning or return a clear config error when tor.proxy is set and use_integrated=true?

impls/src/tor/arti.rs
ardocrat: changed to unwrap_or(false) + logging what we are launching

impls/src/tor/arti.rs
ardocrat: added logging

PR #755, opened 2026-06-03
Update lmdb

ardocrat:
- Update lmdb from `grin_store` crate
- Migrate single backend from trait to struct (since we are using lifetime for iterator and can not use dyn)
- Update tests
- Update cargo deps
- Update design docs
- Do not panic on backend errors
- Store slate state at db (can be needed to resend tx to get slate from slatepack file)


libwallet/src/backend.rs
```diff
@@ -0,0 +1,779 @@
...
+				&mut v,
+				protocol_version,
+				ser::DeserializationMode::default(),
+			)
+			.map_err(From::from)
+		});
+		let iter = prefix_iter
+			.expect("deserialize")
+			.into_iter()
+			.filter(|x| x.is_ok())
```
wiesche89: We should not discard deserialization errors with .filter(|x| x.is_ok()) (see also line 354). Otherwise the wallet can continue with incomplete state.
One possible fix is to change these methods to return Result<Vec<_>, Error>. This materializes all entries, the performance impact would need to be measured. 
But a Result return is needed for clean error handling. panic or expect would also be problematic for CLI/REST usage, i think?

This is a bit more effort, but it handles the error cleanly.

libwallet/src/backend.rs
```diff
@@ -0,0 +1,779 @@
...
+	/// Create new wallet backend.
+	pub fn new(data_file_dir: &str, n_client: C) -> Result<Self, Error> {
+		let db_path = Path::new(data_file_dir).join(DB_DIR);
+		fs::create_dir_all(&db_path).expect("Couldn't create wallet backend directory!");
+
+		let stored_tx_path = Path::new(data_file_dir).join(TX_SAVE_DIR);
+		fs::create_dir_all(&stored_tx_path)
+			.expect("Couldn't create wallet backend tx storage directory!");
+
+		let store = Store::new(
```
wiesche89, line 130: the generic store migration is covered, but this wallet backend change should also have a wallet regression test that opens and migrates an old single lmdb.

libwallet/src/backend.rs
ardocrat: Yes, I guess so too

libwallet/src/backend.rs
ardocrat: agree, we need to handle this error, not just panic (this case if more for CLI)

libwallet/src/backend.rs
```diff
@@ -584,11 +575,8 @@ where
 			)
 			.map_err(From::from)
 		});
-		let iter = prefix_iter?
-			.into_iter()
-			.filter(|x| x.is_ok())
-			.map(|x| x.unwrap());
-		Ok(iter)
+		let items: Vec<OutputData> = prefix_iter?.collect::<Result<Vec<_>, _>>()?;
```
ardocrat, line 580: Here we collect 10k max items then convert them to iter if needed, I guess we can return `Item = Result<OutputData>` here, then handle an error inside iter, missing outputs will require wallet scan

libwallet/src/backend.rs
```diff
@@ -668,11 +656,8 @@ where
 			)
 			.map_err(From::from)
 		});
-		let iter = prefix_iter?
-			.into_iter()
-			.filter(|x| x.is_ok())
-			.map(|x| x.unwrap());
-		Ok(iter)
+		let items: Vec<TxLogEntry> = prefix_iter?.collect::<Result<Vec<_>, _>>()?;
```
ardocrat: If we'll return `Item=Result<TxLogEntry, Error>` can we allow to skip missing txs from wallet history and just log an error?

libwallet/src/backend.rs
```diff
@@ -717,11 +702,8 @@ where
 				)
 				.map_err(From::from)
 			});
-		let iter = prefix_iter?
-			.into_iter()
-			.filter(|x| x.is_ok())
-			.map(|x| x.unwrap());
-		Ok(iter)
+		let items: Vec<AcctPathMapping> = prefix_iter?.collect::<Result<Vec<_>, _>>()?;
```
ardocrat, line 709: I think we can return `Item=Result<AcctPathMapping, Error>` and require wallet rescan if at least one record can not be read

libwallet/src/backend.rs
wiesche89: Yes, returning Item = Result<OutputData, Error> sounds good to me.
The only thing I would add is that the underlying node store iterator error propagation needs to be fixed first, otherwise the wallet side can still miss the real read error.

libwallet/src/backend.rs
wiesche89: I would not skip tx log read errors. The tx log is used for more than display => needs repair => local repair?

libwallet/src/backend.rs
wiesche89: Returning Item = Result<AcctPathMapping, Error> sounds good, but I would not call this a normal rescan case. Account mappings are wallet ocal data, so if one cannot be read we should show the error and require repair or restore instead of continuing silently.

libwallet/src/backend.rs
ardocrat: after rescan tx logs will disappear, it will show only unspent outputs as we create new tx logs only when new tx is appearing

libwallet/src/api_impl/owner.rs
wiesche89: This seems to come from #748 rather than this PR, but it is still in this branch. On an exact batch boundary the last block can be missed, for example start_height = 0 and tip = 10001 only scans up to 10000. I think both scan loops should include the tip height and have a small boundary test for this case.

wiesche89: Same issue here, this loop should also include the tip height, otherwise the last block can be missed on an exact batch boundary.

ardocrat: fixed

ardocrat: fixed

libwallet/src/backend.rs
ardocrat: since there is no pagination anyway left it as you added, collecting all values and returning error if needed

libwallet/src/backend.rs
ardocrat: show errors at log now, not collecting all records at iterator

libwallet/src/backend.rs
ardocrat: left it as you added, there is no pagination at code, so will collect all values to return general error if at least one output can not be read

libwallet/src/api_impl/foreign.rs
```diff
@@ -231,3 +236,29 @@ where
...
+{
+	let tx = wallet
+		.tx_log_iter()?
+		.filter(|tx| tx.is_ok())
+		.map(|tx| tx.unwrap())
+		.find(|tx| tx.tx_slate_id == Some(slate.id));
+	if let Some(mut tx) = tx {
+		let mut batch = wallet.batch(keychain_mask)?;
+		tx.tx_slate_state = Some(slate.state.clone());
+		batch.save_tx_log_entry(tx.clone(), parent_key_id)?;
```
wiesche89, line 265: batch.commit() seems to be missing here, so this slate state update is written to the batch but never persisted. Other batch writes in this file call commit()

libwallet/src/api_impl/foreign.rs
```diff
@@ -231,3 +236,29 @@ where
...
+	parent_key_id: &Identifier,
+	slate: &Slate,
+) -> Result<(), Error>
+where
+	C: NodeClient,
+	K: Keychain,
+{
+	let tx = wallet
+		.tx_log_iter()?
+		.filter(|tx| tx.is_ok())
```
wiesche89: I agree for tx history listing/pagination, especially with the lmbd sorting TODO (Telegram msg). But update_tx_slate_state is a state update path, so I would not filter iterator errors here, a tx log read error should be returned instead of becoming “slate not found”.

libwallet/src/internal/updater.rs
```diff
@@ -94,192 +91,200 @@ where
...
-					if v {
-						tx_entry.tx_type == TxLogEntryType::TxReceived
-							|| tx_entry.tx_type == TxLogEntryType::TxReceivedCancelled
-					} else {
-						true
-					}
+	let mut bad_records = 0;
+	let txs_iter = wallet
+		.tx_log_iter()?
+		.filter(|tx| {
```
wiesche89, line 107: We should surface this error somehow, not only log it. Otherwise callers may trust an incomplete tx history.

libwallet/src/api_impl/foreign.rs
wiesche89: I would prefer that the deserialization error is not swallowed here. Otherwise we cannot distinguish between “slate not found” and “tx-log record could not be deserialized”

libwallet/src/internal/updater.rs
ardocrat: we can not recover tx history, wallet rescan can only restore missing outputs to show right balance

libwallet/src/api_impl/foreign.rs
ardocrat: added amount of bad records at error to have some idea why tx slate can not be found

PR #762, opened 2026-06-21
Global Tor client state

ardocrat:
- Use global Arti client state to reuse on requests when listener is active
- Added logging when start client to send requests
- Do not use cached client to send requests

### DIFF
--- a/impls/src/tor/arti.rs
+++ b/impls/src/tor/arti.rs
@@ -52,6 +52,8 @@ use tor_rtcompat::{SleepProviderExt, ToplevelBlockOn};
 lazy_static! {
 	pub static ref ARTI_RUNTIME: LazyLock<Mutex<Option<ArtiRuntimeWrapper>>> =
 		LazyLock::new(|| Mutex::new(ArtiRuntimeWrapper::create().ok()));
+	pub static ref ARTI_CLIENT_CONFIG: LazyLock<Mutex<Option<(Arc<TorClient<TokioNativeTlsRuntime>>, TorClientConfig)>>> =
+		LazyLock::new(|| Mutex::new(None));
 }
 
 /// Get Tor client runtime.
@@ -59,7 +61,7 @@ fn runtime() -> Result<TokioNativeTlsRuntime, Error> {
 	let mut runtime = ARTI_RUNTIME.lock().unwrap();
 	let r = match runtime.as_ref() {
 		None => runtime.insert(ArtiRuntimeWrapper::create()?),
-		Some(r) => r
+		Some(r) => r,
 	};
 	Ok(r.runtime.clone())
 }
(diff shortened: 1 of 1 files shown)


impls/src/tor/arti.rs
```diff
@@ -238,6 +247,14 @@ fn init_client(
 	cache_path: &PathBuf,
 	config: TorConfig,
 ) -> Result<(Arc<TorClient<TokioNativeTlsRuntime>>, TorClientConfig), Error> {
+	// Return existing client if exists.
+	{
+		let client_config = ARTI_CLIENT_CONFIG.lock().unwrap();
+		if let Some((client, config)) = client_config.as_ref() {
```
wiesche89: I think this cache needs to be keyed or validated against the requested state_path, cache_path, and TorConfig. Right now the first successful client wins for the whole process, so later calls with a different bridge/proxy config or Tor directory silently reuse the old client.

impls/src/tor/arti.rs
wiesche89, line 91: This assumes the returned client was built from the same state_path, but with the global cache that may no longer be true. We could end up writing the hidden-service key into a keystore that the cached client does not use. Can we make sure the client and key path always come from the same effective config?

impls/src/tor/arti.rs
```diff
@@ -238,6 +247,14 @@ fn init_client(
 	cache_path: &PathBuf,
 	config: TorConfig,
 ) -> Result<(Arc<TorClient<TokioNativeTlsRuntime>>, TorClientConfig), Error> {
+	// Return existing client if exists.
+	{
+		let client_config = ARTI_CLIENT_CONFIG.lock().unwrap();
```
wiesche89: One concern here, if two callers hit this at startup, they can both see an empty cache and start bootstrapping in parallel. Then the last one to finish wins and replaces the global client. Could we make the first initialization single-flight, so the second caller waits for the first result instead of starting its own bootstrap?

impls/src/tor/arti.rs
```diff
@@ -238,6 +247,14 @@ fn init_client(
...
+		if let Some((client, config)) = client_config.as_ref() {
+			return Ok((client.clone(), config.clone()));
```
wiesche89: Could we add a debug log when the cached client is reused? If a later Tor config change does not take effect, it would otherwise be hard to tell from the logs that the old client is still being used

impls/src/tor/arti.rs
ardocrat: it means current service must be restarted with new key on `set_active_account` and `set_tor_config` currently this behaviour is not handled at all, also for current process.

to simplify process I will compare configs on client initialization, new config can be built, as we also need to compare if bridge was changed

impls/src/tor/arti.rs
ardocrat: it means current service must be restarted with new key on `set_active_account` and `set_tor_config` currently this behaviour is not handled at all, also for current process

impls/src/tor/arti.rs
ardocrat: I will keep lifetime of `ARTI_CLIENT_CONFIG` locked till end of this function to write result in the end, then we can make sure another call will see result.

impls/src/tor/arti.rs
ardocrat: Added logging for client reuse or reinit.

impls/src/tor/arti.rs
```diff
@@ -272,6 +281,18 @@ fn init_client(
 		.build()
 		.map_err(|e| Error::TorConfig(format!("{:?}", e)))?;
 
+	// Return existing client if exists and config was not changed.
+	let mut client_config = ARTI_CLIENT_CONFIG.lock().unwrap();
```
wiesche89: This lock is held while Tor is bootstrapping, which can take a while.
In interactive CLI mode, listen can bootstrap in the background and send can run right after. In that case send waits here before it can start its own client.  Follow up?

ardocrat: Makes sense, we currently use separate tor state path for sender, also listener can be not active so we need to stop client after command execution

PR #763, opened 2026-06-22
Optimize tor sync flow

ardocrat:
- Return concrete error on sync workflow instead of `None`
- Update slate state after finalize over Tor
- Output slate to file after finalize request

### DIFF
--- a/api/src/foreign.rs
+++ b/api/src/foreign.rs
@@ -373,7 +373,7 @@ where
 			Some(a) => {
 				let tc = self.tor_config.lock();
 				let can_send = if let Some(tc) = tc.as_ref() {
-					tc.skip_send(None)
+					tc.send_tor(None)
 				} else {
 					false
 				};
--- a/api/src/foreign.rs
+++ b/api/src/foreign.rs
@@ -24,6 +24,7 @@ use crate::libwallet::{
 use crate::try_slatepack_sync_workflow;
 use crate::util::secp::key::SecretKey;
 use crate::util::Mutex;
+use libwallet::api_impl::types::update_tx_slate_state;
 use std::sync::Arc;
 
 /// ForeignAPI Middleware Check callback
@@ -381,7 +382,16 @@ where
 				}
 				let res = try_slatepack_sync_workflow(&ret_slate, &a, tc.clone(), None, true);
 				match res {
-					Ok(s) => Ok(s),
+					Ok(s) => {
+						let parent_key_id = w.parent_key_id();
+						let _ = update_tx_slate_state(
+							w,
+							(&self.keychain_mask).as_ref(),
+							&parent_key_id,
+							&s,
+						);
+						Ok(s)
+					}
 					Err(e) => {
 						error!("Error on sending over Tor: {}", e);
 						Ok(ret_slate)
--- a/api/src/foreign.rs
+++ b/api/src/foreign.rs
@@ -384,12 +384,15 @@ where
 				match res {
 					Ok(s) => {
 						let parent_key_id = w.parent_key_id();
-						let _ = update_tx_slate_state(
+						match update_tx_slate_state(
 							w,
 							(&self.keychain_mask).as_ref(),
 							&parent_key_id,
 							&s,
-						);
+						) {
+							Ok(_) => {}
+							Err(e) => error!("Error on updating slate state: {}", e),
+						}
 						Ok(s)
 					}
 					Err(e) => {
--- a/api/src/foreign.rs
+++ b/api/src/foreign.rs
@@ -370,18 +370,22 @@ where
 		)?;
 		match r_addr {
 			Some(a) => {
-				let tor_config_lock = self.tor_config.lock();
-				let res = try_slatepack_sync_workflow(
-					&ret_slate,
-					&a,
-					tor_config_lock.clone(),
-					None,
-					true,
-					self.doctest_mode,
-				);
+				let tc = self.tor_config.lock();
+				let can_send = if let Some(tc) = tc.as_ref() {
+					tc.skip_send(None)
+				} else {
+					false
+				};
+				if self.doctest_mode || !can_send {
+					return Ok(ret_slate);
+				}
+				let res = try_slatepack_sync_workflow(&ret_slate, &a, tc.clone(), None, true);
 				match res {
-					Ok(s) => Ok(s.unwrap()),
-					Err(_) => Ok(ret_slate),
+					Ok(s) => Ok(s),
+					Err(e) => {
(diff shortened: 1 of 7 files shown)


config/src/types.rs
```diff
@@ -194,6 +194,16 @@ impl Default for TorConfig {
 	}
 }
 
+impl TorConfig {
+	/// Check if attempt to send over Tor must be skipped using provided possible argument at priority.
+	pub fn skip_send(&self, skip_arg: Option<bool>) -> bool {
```
wiesche89: I think this helper is being used with inverted semantics. skip_send() returns whether we should skip Tor, but the call sites treat it as can_send.

With the default config this skips Tor sending, while --manual actually tries to send. I’d rename or rework this into positive should_send logic, or flip the checks at the call sites.

controller/src/command.rs
```diff
@@ -431,18 +431,37 @@ where
...
+			&slate,
+			args.dest.as_str(),
+			args.outfile,
+			true,
+			false,
+			args.slatepack_qr,
+		)?)
+	};
+
+	let can_send = if let Some(tc) = tor_config.as_ref() {
```
wiesche89, line 453: I think this boolean is flipped. skip_send() returns whether the Tor send should be skipped, but the result is used as can_send here. With the default config this prevents Tor from being attempted, while --manual or skip_tor = true actually enables the Tor send. I’d rename this to should_skip and invert the condition at all call sites.

api/src/owner.rs
```diff
@@ -673,28 +676,20 @@ where
...
 				let tc = tor_config_lock.clone();
-				let tc = match tc {
-					Some(mut c) => {
-						if let Some(skip_tor) = sa.skip_tor {
-							c.skip_send_attempt = Some(skip_tor);
-						}
-						Some(c)
-					}
-					None => None,
+				let can_send = if let Some(tc) = tc.as_ref() {
```
wiesche89, line 679: See comment controller/src/command.rs line 452

api/src/owner.rs
```diff
@@ -841,26 +836,26 @@ where
...
+					return Ok(slate);
+				}
+				let res = try_slatepack_sync_workflow(&slate, &sa.dest, tc, None, true);
 				match res {
-					Ok(s) => Ok(s.unwrap()),
-					Err(_) => Ok(slate),
+					Ok(s) => {
+						let parent_key_id = w.parent_key_id();
+						let _ = update_tx_slate_state(w, keychain_mask, &parent_key_id, &s);
+						let _ = output_slatepack_file(&self, keychain_mask, &s, &sa.dest);
```
wiesche89: This looks like it can deadlock. process_invoice_tx still holds wallet_inst.lock() when output_slatepack_file() is called, and that path calls create_slatepack_message(), which locks the same wallet_inst again via get_slatepack_address(). Since this mutex is not reentrant, the call can block forever. I think the wallet lock should be scoped and released before the Tor send / output_slatepack_file path.

api/src/owner.rs
```diff
@@ -841,26 +836,26 @@ where
...
+				if self.doctest_mode || !can_send {
+					return Ok(slate);
+				}
+				let res = try_slatepack_sync_workflow(&slate, &sa.dest, tc, None, true);
 				match res {
-					Ok(s) => Ok(s.unwrap()),
-					Err(_) => Ok(slate),
+					Ok(s) => {
+						let parent_key_id = w.parent_key_id();
+						let _ = update_tx_slate_state(w, keychain_mask, &parent_key_id, &s);
```
wiesche89: Should we at least log this error? If update_tx_slate_state() fails, the caller still gets Ok even though the local tx slate state was not updated after a successful Tor roundtrip. That could leave the wallet state misleading after restart or when shown in tx_list.

api/src/foreign.rs
```diff
@@ -370,18 +371,31 @@ where
...
+				if self.doctest_mode || !can_send {
+					return Ok(ret_slate);
+				}
+				let res = try_slatepack_sync_workflow(&ret_slate, &a, tc.clone(), None, true);
 				match res {
-					Ok(s) => Ok(s.unwrap()),
-					Err(_) => Ok(ret_slate),
+					Ok(s) => {
+						let parent_key_id = w.parent_key_id();
+						let _ = update_tx_slate_state(
```
wiesche89: Should we at least log this error? If update_tx_slate_state() fails, the caller still gets Ok even though the local tx slate state was not updated after a successful Tor roundtrip. That could leave the wallet state misleading after restart or when shown in tx_list.

api/src/owner.rs
wiesche89: if writing the Slatepack file is best effort, I think failures should still be logged. Otherwise File::create , write_all , sync_all errors are silently ignored and the caller has no way to know the expected fallback file was not written.

config/src/types.rs
ardocrat: Inversed

controller/src/command.rs
ardocrat: Renamed skip_send to send_tor

api/src/owner.rs
ardocrat: Fixed

api/src/owner.rs
ardocrat: fixed with scope

ardocrat: Added errors logging

api/src/foreign.rs
ardocrat: Added errors logging

api/src/owner.rs
ardocrat: Added errors logging

controller/src/command.rs
```diff
@@ -686,33 +693,45 @@ where
...
+	};
+	if test_mode || !can_send {
+		return output_sp();
+	}
+
+	let res = try_slatepack_sync_workflow(&slate, &dest, tor_config, None, true);
 
 	match res {
-		Ok(Some(_)) => {
+		Ok(_) => {
```
wiesche89: Should we use the returned slate here as well? The API path now updates the local tx_slate_state after a successful Tor roundtrip, but the CLI receive path still ignores Ok(s). That means receive via API can move S2 -> S3 locally, while receive via CLI only prints success. Same question applies to the process_invoice path below for I2 -> I3.

controller/src/command.rs
```diff
@@ -1012,41 +1031,50 @@ where
...
```
wiesche89: Same question as above: should the CLI update the local tx_slate_state from the returned final slate here too? The API path now records I2 -> I3 after a successful Tor roundtrip, but this CLI path ignores Ok(s).

PR #765, opened 2026-06-26
Connection timeout

ardocrat:
- Use default node API connection timeout from config
- Use Tor boostrap and request timeout from config

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -76,7 +76,7 @@ fn comments() -> HashMap<String, String> {
 	retval.insert(
 		"node_api_request_timeout_secs".to_string(),
 		"
-#connection timeout for requests to node API
+#total timeout for requests to node API
 "
 		.to_string(),
 	);
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -74,6 +74,13 @@ fn comments() -> HashMap<String, String> {
 		.to_string(),
 	);
 	retval.insert(
+		"node_api_connection_timeout_secs".to_string(),
+		"
+#connection timeout for requests to node API
+"
+		.to_string(),
+	);
+	retval.insert(
 		"node_api_secret_path".to_string(),
 		"
 #location of the node api secret for basic auth on the Grin API
@@ -248,6 +255,22 @@ fn comments() -> HashMap<String, String> {
 		"send_config_dir".to_string(),
 		"
 #Directory to output TOR configuration to when sending
+"
+		.to_string(),
+	);
+
+	retval.insert(
+		"request_timeout_secs".to_string(),
+		"
+#Tor request timeout in seconds
+"
+		.to_string(),
+	);
+
+	retval.insert(
+		"boostrap_timeout_secs".to_string(),
+		"
+#Tor boostrap timeout in seconds
 "
 		.to_string(),
 	);
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -270,7 +270,7 @@ fn comments() -> HashMap<String, String> {
 	retval.insert(
 		"bootstrap_timeout_secs".to_string(),
 		"
-#Tor boostrap timeout in seconds
+#Tor bootstrap timeout in seconds
 "
 		.to_string(),
 	);
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -268,7 +268,7 @@ fn comments() -> HashMap<String, String> {
 	);
 
 	retval.insert(
-		"boostrap_timeout_secs".to_string(),
+		"bootstrap_timeout_secs".to_string(),
 		"
 #Tor boostrap timeout in seconds
 "
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -74,7 +74,7 @@ fn comments() -> HashMap<String, String> {
 		.to_string(),
 	);
 	retval.insert(
-		"node_api_connection_timeout_secs".to_string(),
+		"node_api_request_timeout_secs".to_string(),
 		"
 #connection timeout for requests to node API
 "
--- a/config/src/types.rs
+++ b/config/src/types.rs
@@ -181,7 +181,7 @@ pub struct TorConfig {
 	/// Connection timeout for request in seconds.
 	pub request_timeout_secs: Option<u64>,
 	/// Connection timeout for boostrap in seconds.
-	pub boostrap_timeout_secs: Option<u64>,
+	pub bootstrap_timeout_secs: Option<u64>,
 	/// Tor bridge config
 	#[serde(default)]
 	pub bridge: TorBridgeConfig,
@@ -199,7 +199,7 @@ impl Default for TorConfig {
 			socks_proxy_addr: "127.0.0.1:59050".to_owned(),
 			send_config_dir: ".".into(),
 			request_timeout_secs: Some(Self::REQUEST_TIMEOUT_SECS),
-			boostrap_timeout_secs: Some(Self::BOOTSTRAP_TIMEOUT_SECS),
+			bootstrap_timeout_secs: Some(Self::BOOTSTRAP_TIMEOUT_SECS),
 			bridge: TorBridgeConfig::default(),
 			proxy: TorProxyConfig::default(),
 		}
--- a/config/src/types.rs
+++ b/config/src/types.rs
@@ -119,6 +119,7 @@ impl WalletConfig {
 	pub fn api_request_timeout(&self) -> Duration {
 		Duration::from_secs(
 			self.node_api_request_timeout_secs
+				.filter(|s| *s > 0)
 				.unwrap_or(Self::NODE_API_REQUEST_TIMEOUT_SECS),
 		)
 	}
@@ -232,6 +233,7 @@ impl TorConfig {
 	pub fn request_timeout(&self) -> Duration {
 		Duration::from_secs(
 			self.request_timeout_secs
+				.filter(|s| *s > 0)
 				.unwrap_or(Self::REQUEST_TIMEOUT_SECS),
 		)
 	}
@@ -240,6 +242,7 @@ impl TorConfig {
 	pub fn bootstrap_timeout(&self) -> Duration {
 		Duration::from_secs(
 			self.bootstrap_timeout_secs
+				.filter(|s| *s > 0)
 				.unwrap_or(Self::BOOTSTRAP_TIMEOUT_SECS),
 		)
 	}
--- a/config/src/types.rs
+++ b/config/src/types.rs
@@ -14,12 +14,12 @@
 
 //! Public types for config modules
 
+use crate::core::global::ChainTypes;
+use crate::util::logger::LoggingConfig;
 use std::fmt;
 use std::io;
 use std::path::PathBuf;
-
-use crate::core::global::ChainTypes;
-use crate::util::logger::LoggingConfig;
+use std::time::Duration;
 
 /// Command-line wallet configuration
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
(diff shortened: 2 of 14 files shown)


config/src/types.rs
```diff
@@ -172,6 +178,10 @@ pub struct TorConfig {
 	pub socks_proxy_addr: String,
 	/// Send configuration directory
 	pub send_config_dir: String,
+	/// Connection timeout for request in seconds.
+	pub request_timeout_secs: Option<u64>,
+	/// Connection timeout for boostrap in seconds.
+	pub boostrap_timeout_secs: Option<u64>,
```
wiesche89: Typo in the new config key: boostrap_timeout_secs should probably be bootstrap_timeout_secs before this becomes part of the public config.

impls/src/adapters/tor.rs
```diff
@@ -183,14 +183,19 @@ impl TorSlateSender {
 				SocketAddr::V4(self.config.socks_proxy_addr.parse().map_err(|_| {
 					ClientError::Internal("Socks proxy address is not set".to_string())
 				})?);
-			let client = Client::with_proxy(socks_proxy_addr, "socks5h://")
+			let timeout = Duration::from_secs(
+				self.config
+					.request_timeout_secs
```
wiesche89: We now use request_timeout_secs for the external Tor request, but the external Tor process bootstrap timeout above still stays hardcoded at 20s. Should that use the new bootstrap timeout config too?

impls/src/node_clients/http.rs
```diff
@@ -45,25 +45,27 @@ impl HTTPNodeClient {
 	pub fn new(
 		node_url: &str,
 		node_api_secret: Option<String>,
+		request_timeout: Option<Duration>,
```
wiesche89: Since this changes the HTTPNodeClient::new signature, we should also update the excluded integration crate call sites. They still call this with two args and would fail once those tests are built.

impls/src/client_utils/client.rs
```diff
@@ -52,22 +53,28 @@ pub struct Client {
...
-	fn build(proxy: Option<(SocketAddr, &str)>) -> Result<Self, Error> {
+	fn build(proxy: Option<(SocketAddr, &str)>, timeout: Option<Duration>) -> Result<Self, Error> {
 		let mut headers = HeaderMap::new();
 		headers.insert(USER_AGENT, HeaderValue::from_static("grin-client"));
 		headers.insert(ACCEPT, HeaderValue::from_static("application/json"));
 		headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
 
 		let mut builder = ClientBuilder::new()
-			.timeout(Duration::from_secs(20))
+			.timeout(timeout.unwrap_or_else(|| {
```
wiesche89: reqwest::ClientBuilder::timeout is a total request timeout, not just a connection timeout. Maybe rename/comment this as request timeout to avoid confusion.

This makes the generic HTTP client use a wallet specific timeout default. Would it be cleaner to keep a local default here, or make callers always pass the resolved timeout?

config/src/types.rs
```diff
@@ -188,13 +198,20 @@ impl Default for TorConfig {
...
+			boostrap_timeout_secs: Some(Self::BOOTSTRAP_TIMEOUT_SECS),
 			bridge: TorBridgeConfig::default(),
 			proxy: TorProxyConfig::default(),
 		}
 	}
 }
 
 impl TorConfig {
+	/// Tor request timeout in seconds.
+	pub const REQUEST_TIMEOUT_SECS: u64 = 60;
```
wiesche89: The timeout fallback logic is now repeated in several places. Maybe add small helpers on the config types, e.g. request_timeout() and bootstrap_timeout(), so the default + Duration::from_secs logic lives in one place.

config/src/types.rs
ardocrat: Thx, fixed

impls/src/adapters/tor.rs
ardocrat: FIxed

impls/src/node_clients/http.rs
ardocrat: Updated, Integration tests are still outdated, need to fix them at separate PR

impls/src/client_utils/client.rs
ardocrat: This is exactly node http client (config comment: `connection timeout for requests to node API`) that can be used as proxy for Tor (we also use `request_timeout_secs` variable here), renamed variables at client and config.

config/src/types.rs
ardocrat: Added functions to config

controller/src/controller.rs
```diff
@@ -88,27 +87,25 @@ fn init_tor_listener(
...
 	tor_dir: String,
 	addr: &str,
-	bridge: TorBridgeConfig,
-	tor_proxy: TorProxyConfig,
+	tor_config: TorConfig,
 ) -> Result<Tor, Error> {
 	info!("Starting external Tor Process listener.");
 
 	let mut process = tor_process::TorProcess::new();
+	let tor_timeout = tor_config.bootstrap_timeout().as_secs();
```
wiesche89, line 95: Can we keep the previous bridge buffer? Bridges intentionally had a higher timeout before (40 vs 20). With a low configured value we lose that margin a max(40) in the bridge case would cover it.

Or do you think that's enough?

config/src/types.rs
```diff
@@ -108,6 +114,14 @@ impl WalletConfig {
 		self.accept_fee_base
 			.unwrap_or_else(|| WalletConfig::default_accept_fee_base())
 	}
+
+	/// Node API requests timeout.
+	pub fn api_request_timeout(&self) -> Duration {
+		Duration::from_secs(
```
wiesche89, line 120: A small lower bound would help with 0 from config this path fails immediately and the error doesn’t point back to the config. filter(|s| *s > 0).unwrap_or(default) would handle it cleanly.

impls/src/client_utils/client.rs
```diff
@@ -52,22 +53,31 @@ pub struct Client {
...
+	) -> Result<Self, Error> {
 		let mut headers = HeaderMap::new();
 		headers.insert(USER_AGENT, HeaderValue::from_static("grin-client"));
 		headers.insert(ACCEPT, HeaderValue::from_static("application/json"));
 		headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
 
 		let mut builder = ClientBuilder::new()
-			.timeout(Duration::from_secs(20))
+			.timeout(request_timeout.unwrap_or_else(|| {
+				Duration::from_secs(WalletConfig::NODE_API_REQUEST_TIMEOUT_SECS)
```
wiesche89: I’m still a bit unsure about using WalletConfig as the fallback here. Since this generic client is also used for the Tor socks path, a local default or a caller-provided resolved timeout would keep the layer boundary cleaner.

config/src/types.rs
```diff
@@ -37,6 +37,8 @@ pub struct WalletConfig {
 	/// The api address of a running server node against which transaction inputs
 	/// will be checked during send
 	pub check_node_api_http_addr: String,
+	/// Connection timeout for requests to node API.
```
wiesche89: The code is fine, but the comment still says connection timeout. This is really a total request timeout (connect + response), so after the rename to request_timeout we should align the comments.

src/bin/grin-wallet.rs
```diff
@@ -26,10 +26,12 @@ use grin_core as core;
 use grin_util as util;
 use grin_wallet::cmd;
 use grin_wallet_config as config;
+use grin_wallet_config::WalletConfig;
```
wiesche89: Small cleanup request, this import is unused and now adds a cargo-check warning.

config/src/types.rs
```diff
@@ -172,6 +186,10 @@ pub struct TorConfig {
 	pub socks_proxy_addr: String,
 	/// Send configuration directory
 	pub send_config_dir: String,
+	/// Connection timeout for request in seconds.
+	pub request_timeout_secs: Option<u64>,
+	/// Connection timeout for boostrap in seconds.
```
wiesche89: Small typo cleanup
boostrap -> bootstrap

config/src/types.rs
```diff
@@ -188,20 +206,43 @@ impl Default for TorConfig {
...
 			bridge: TorBridgeConfig::default(),
 			proxy: TorProxyConfig::default(),
 		}
 	}
 }
 
 impl TorConfig {
+	/// Tor request timeout in seconds.
+	const REQUEST_TIMEOUT_SECS: u64 = 60;
+	/// Tor boostrap timeout in seconds.
```
wiesche89: Small typo cleanup
boostrap -> bootstrap

config/src/comments.rs
```diff
@@ -252,6 +259,22 @@ fn comments() -> HashMap<String, String> {
...
+		"
+#Tor request timeout in seconds
+"
+		.to_string(),
+	);
+
+	retval.insert(
+		"bootstrap_timeout_secs".to_string(),
+		"
+#Tor boostrap timeout in seconds
```
wiesche89: Small typo cleanup
boostrap -> bootstrap

controller/src/controller.rs
ardocrat: Must be enough with default timeout of 60s

config/src/comments.rs
ardocrat: Fixed

config/src/types.rs
ardocrat: Fixed

config/src/types.rs
ardocrat: Fixed

config/src/types.rs
ardocrat: Fixed

src/bin/grin-wallet.rs
ardocrat: Fixed

impls/src/client_utils/client.rs
ardocrat: We should make argument non-optional, cause we always provide it from config

controller/src/controller.rs
ardocrat: From my test bridge connection took 10s with good connection

config/src/types.rs
ardocrat: Added

impls/src/tor/process.rs
```diff
@@ -137,7 +137,7 @@ impl TorProcess {
 		self
 	}
 
-	pub fn timeout(&mut self, timeout: u32) -> &mut Self {
+	pub fn timeout(&mut self, timeout: u64) -> &mut Self {
```
wiesche89, line 140: Can we add an upper bound or validation here? The timeout now comes from config as a u64, but launch() later casts it to i64 for chrono::Duration::seconds().

config/src/comments.rs
```diff
@@ -70,6 +70,13 @@ fn comments() -> HashMap<String, String> {
 		"check_node_api_http_addr".to_string(),
 		"
 #where the wallet should find a running node
+"
+		.to_string(),
+	);
+	retval.insert(
+		"node_api_request_timeout_secs".to_string(),
+		"
+#connection timeout for requests to node API
```
wiesche89: This says connection timeout, but this is really the HTTP client's total request timeout. Could we rename this to request timeout and mention that 0 falls back to the default?

config/src/types.rs
```diff
@@ -172,6 +187,10 @@ pub struct TorConfig {
 	pub socks_proxy_addr: String,
 	/// Send configuration directory
 	pub send_config_dir: String,
+	/// Connection timeout for request in seconds.
```
wiesche89: Small doc cleanup, these are not pure connection timeouts. Request timeout and Bootstrap timeout would match the actual behavior and the field names better.

impls/src/tor/process.rs
ardocrat: added check for `i64::MAX`

config/src/comments.rs
ardocrat: fixed comment

config/src/types.rs
ardocrat: fixed doc

PR #768, opened 2026-07-03
Update deps

ardocrat:
Update common dependencies to fix possible security vulnerabilities and remove unused.

- Update grin submodule to last staging release
- Remove `remove_dir_all` as main Windows issue fixed with std function: https://github.com/rust-lang/rust/issues/29497, so there is no problem with CI https://github.com/mimblewimble/grin-wallet/pull/604 
- Update `rpassword` (closes https://github.com/mimblewimble/grin-wallet/pull/761)
- Update `semver` from `0.10` to `1.0.28`
- Update `ring` from `0.16` to `0.17.4`
- Update `rustyline` from `6` to `9`
- Update `ed25519-dalek` from `1` to `2`
- Update `curve25519-dalek` from `2` to `5`
- Update `age` from `0.7` to `0.8`
- Update `uuid` from `0.8` to `1.23.4`
- Update `x25519-dalek` from `0.6` to `2`
- Update `arti` from `0.43` to `0.44`
- Update `grin` submodule
- Update `hyper` to `1.10.1`
- Remove unused dependencies


tests/owner_v3_init_secure.rs
```diff
@@ -41,6 +41,7 @@ use common::{
 
 #[test]
 fn owner_v3_init_secure() -> Result<(), grin_wallet_controller::Error> {
+	rustls::crypto::ring::default_provider().install_default().expect("Failed to install rustls crypto provider");
```
wiesche89: This fixes the rustls provider issue for this test, but the same initialization seems to be missing in the production path. With the updated dependency graph, the TLS API listener can still reach ServerConfig::builder() without a default provider installed. Can we install rustls::crypto::ring::default_provider().install_default() once early in real_main()?

impls/src/tor/config.rs
```diff
@@ -101,9 +101,11 @@ pub fn create_onion_service_sec_key_file(
 	// Tag is always 32 bytes, so pad with null zeroes
 	file.write(b"== ed25519v1-secret: type0 ==\0\0\0")
 		.map_err(|_| Error::IO)?;
-	let expanded_skey: ExpandedSecretKey = ExpandedSecretKey::from(sec_key);
-	file.write_all(&expanded_skey.to_bytes())
-		.map_err(|_| Error::IO)?;
+	let expanded_skey: ExpandedSecretKey = ExpandedSecretKey::from(sec_key.as_bytes());
+	let mut sk_bytes = [0_u8; 64];
+	sk_bytes[0..32].copy_from_slice(&expanded_skey.scalar.to_bytes());
```
wiesche89: This subtly changes the keyfile bytes: dalek 2.x gives us the mod-l-reduced scalar here, while the old code wrote the clamped unreduced value. This should be mathematically equivalent for Tor, but it differs from the old/usual format. Can we either reproduce the old format exactly, or verify this once against a real Tor binary and add a short comment?

src/cmd/wallet.rs
```diff
@@ -42,7 +42,7 @@ where
 
 	// This will also cache the node version info for calls to foreign API check middleware
 	if let Some(v) = node_client.clone().get_version_info() {
-		if Version::parse(&v.node_version) < Version::parse(MIN_COMPAT_NODE_VERSION) {
+		if Version::parse(&v.node_version).ok() < Version::parse(MIN_COMPAT_NODE_VERSION).ok() {
```
wiesche89: None < Some is true. An unparsable node version is now rejected as outdated, while it used to pass. An explicit match with a warning on parse failure would be clearer and avoids a misleading error message.

src/cmd/wallet_args.rs
```diff
@@ -68,7 +68,7 @@ impl From<std::io::Error> for ParseError {
 }
 
 fn prompt_password_stdout(prompt: &str) -> ZeroingString {
```
wiesche89: Small naming cleanup, this wrapper is still called prompt_password_stdout, but it now calls rpassword::prompt_password(). Could we rename it or add a short comment so the stdout behavior is not implied after the dependency update?

libwallet/src/slate_versions/ser.rs
```diff
@@ -151,15 +151,19 @@ pub mod dalek_seckey_serde {
 		String::deserialize(deserializer)
 			.and_then(|string| from_hex(&string).map_err(|err| Error::custom(err.to_string())))
 			.and_then(|bytes: Vec<u8>| {
-				DalekSecretKey::from_bytes(&bytes).map_err(|err| Error::custom(err.to_string()))
+				let bytes = bytes.into_boxed_slice();
```
wiesche89: This into_boxed_slice().as_array() pattern appears in several dalek conversion spots. Could we use <&[u8; 32]>::try_from(&bytes[..]) instead? That avoids the allocation.

libwallet/src/mwixnet/onion/crypto/dalek.rs
```diff
@@ -116,84 +127,12 @@ impl Writeable for DalekPublicKey {
...
-}
-
 #[cfg(test)]
 mod tests {
 	use super::*;
 	use crate::mwixnet::onion::test_util::rand_keypair;
 	use grin_core::ser::{self, ProtocolVersion};
 	use grin_util::ToHex;
-	use rand::Rng;
 	use serde::{Deserialize, Serialize};
```
wiesche89, line 198: Since the signature wrapper and sig_test are removed in this hunk, could we keep a small fixed vector signing/verification test somewhere else? That would preserve coverage across the dalek upgrade without keeping the old wrapper around.

impls/src/tor/config.rs
ardocrat: same format we use at arti (new dalek was here since the beginning), we can share same function to not repeat the code

wiesche89: That makes sense. Sharing the helper sounds good.

tests/owner_v3_init_secure.rs
ardocrat: added

src/cmd/wallet.rs
ardocrat: added logging

src/cmd/wallet_args.rs
ardocrat: renamed

libwallet/src/slate_versions/ser.rs
ardocrat: fixed

libwallet/src/mwixnet/onion/crypto/dalek.rs
ardocrat: fixed tests

src/cmd/wallet_args.rs
```diff
@@ -67,13 +67,13 @@ impl From<std::io::Error> for ParseError {
 	}
 }
 
-fn prompt_password_stdout(prompt: &str) -> ZeroingString {
-	ZeroingString::from(rpassword::prompt_password_stdout(prompt).unwrap())
+fn prompt_password_internal(prompt: &str) -> ZeroingString {
+	ZeroingString::from(rpassword::prompt_password(prompt).unwrap())
```
wiesche89: The rename looks good. One separate concern: prompt_password() no longer reads from stdin. Could this break scripts or headless setups that pipe the password in? Could we also avoid the unwrap() so it fails cleanly?

impls/src/tor/config.rs
```diff
@@ -92,6 +92,13 @@ impl TorRcConfig {
 	}
 }
 
+pub fn exp_sec_key_bytes(expanded_sk: ExpandedSecretKey) -> [u8; 64] {
```
wiesche89: Thanks for sharing this helper. Could we add a short comment saying that this format works with both C Tor and arti? That would make future dependency updates easier to review.

controller/src/controller.rs
```diff
@@ -570,6 +571,8 @@ impl OwnerV3Helpers {
 	}
 }
 
+pub type ApiBody = BoxBody<Bytes, Infallible>;
```
wiesche89: ApiBody is already defined in crate::api. Could we reuse that one here instead of defining it again?

src/cmd/wallet.rs
```diff
@@ -42,10 +42,19 @@ where
 
 	// This will also cache the node version info for calls to foreign API check middleware
 	if let Some(v) = node_client.clone().get_version_info() {
-		if Version::parse(&v.node_version) < Version::parse(MIN_COMPAT_NODE_VERSION) {
-			println!("The Grin Node in use (version {}) is outdated and incompatible with this wallet version.", v.node_version);
+		let node_version = Version::parse(&v.node_version);
+		if let Ok(v) = node_version {
+			if Some(v.clone()) < Version::parse(MIN_COMPAT_NODE_VERSION).ok() {
```
wiesche89: Could we parse the known valid minimum version once and compare v < min_version directly? The Option<Version> ordering and clone make this harder to read. Including the received version string in the parse error would also help diagnostics.

src/bin/grin-wallet.rs
```diff
@@ -131,7 +131,8 @@ fn real_main() -> i32 {
 		},
 	};
 
-	//config.members.as_mut().unwrap().wallet.chain_type = Some(chain_type);
+	// Initialize TLS provider.
+	rustls::crypto::ring::default_provider().install_default().expect("Failed to install rustls crypto provider");
```
wiesche89: This covers the grin-wallet binary. Could we add a short migration note that applications embedding the controller/TLS listener as a library must install a rustls crypto provider themselves?

libwallet/src/mwixnet/onion/crypto/dalek.rs
```diff
@@ -178,22 +182,19 @@ pub mod dalek_sig_serde {
...
-	let sig = keypair.sign(&message);
+	use ed25519_dalek::{SigningKey, Signer};
+	let secret = SigningKey::from_bytes(&sk.0);
+	let sig = secret.sign(&message);
 	Ok(DalekSignature(sig))
 }
 
 #[cfg(test)]
 mod tests {
+	use rand::Rng;
```
wiesche89: Thanks for restoring the test. Could we replace the random values with fixed ones, as originally suggested? That would make it a stable regression test for future dalek updates.

wiesche89 (commented):
cargo fmt --all -- --check currently fails and reports formatting changes in several files. Could you run cargo fmt --all before merging?

src/cmd/wallet_args.rs
ardocrat: tests are passing, I guess it was just rename of method

impls/src/tor/config.rs
ardocrat: added

controller/src/controller.rs
ardocrat: imported

src/cmd/wallet_args.rs
ardocrat: added io error handling

ardocrat:
> cargo fmt --all -- --check currently fails and reports formatting changes in several files. Could you run cargo fmt --all before merging?

for some reason `cargo fmt --all` is not working before commit for `grin-wallet` unlike for `grin` repo, will check this.

src/cmd/wallet.rs
ardocrat: added

src/bin/grin-wallet.rs
ardocrat: I guess this should be set once at `rest.rs` from grin submodule

ardocrat: Will be fixed and removed after merge https://github.com/mimblewimble/grin/pull/3912 to staging

libwallet/src/mwixnet/onion/crypto/dalek.rs
ardocrat: thats just messages, so should be OK, added fixed values.

src/cmd/wallet_args.rs
```diff
@@ -67,25 +67,27 @@ impl From<std::io::Error> for ParseError {
 	}
 }
 
-fn prompt_password_stdout(prompt: &str) -> ZeroingString {
-	ZeroingString::from(rpassword::prompt_password_stdout(prompt).unwrap())
+fn prompt_password_internal(prompt: &str) -> Result<ZeroingString, Error> {
+	Ok(ZeroingString::from(
+		rpassword::prompt_password(prompt).map_err(|e| Error::GenericError(format!("{}", e)))?,
```
wiesche89: Thanks for fixing the error handling. Piped passwords may still be ignored after this update, which could cause existing scripts to wait indefinitely. Could we keep the old behavior or mention that scripts should use -p?

Cargo.toml
```diff
@@ -21,22 +21,20 @@ exclude = ["integration"]
...
 thiserror = "1"
-prettytable-rs = "0.10"
 log = "0.4"
 linefeed = "0.6"
-semver = "0.10"
-rustyline = "6"
+semver = "1.0.28"
+rustyline = "9"
 lazy_static = "1"
+rustls = { version = "0.23.41", default-features = false, features = ["ring"] }
```
wiesche89: It looks like this dependency is no longer used directly after the latest changes. Can we remove it?

libwallet/src/mwixnet/onion/crypto/dalek.rs
```diff
@@ -263,30 +263,30 @@ mod tests {
 	fn sig_test() -> Result<(), Box<dyn std::error::Error>> {
 		// Sign a message
 		let (sk, pk) = rand_keypair();
-		let msg: [u8; 16] = rand::thread_rng().gen();
+		let msg: [u8; 16] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16];
```
wiesche89, line 270: The messages are fixed now, but the keys are still random. Could we use fixed keys as well, so this becomes a stable regression test?

libwallet/src/mwixnet/onion/crypto/dalek.rs
```diff
@@ -263,30 +263,30 @@ mod tests {
...
 		let sig = sign(&sk, &msg).unwrap();
 
 		// Verify signature
 		assert!(sig.verify(&pk, &msg).is_ok());
 
 		// Wrong message
-		let wrong_msg: [u8; 16] = rand::thread_rng().gen();
+		let wrong_msg: [u8; 16] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 17];
```
wiesche89, line 277: Same point

wiesche89 (commented):
The latest changes look good and the previous concerns have been addressed. I only left a few small cleanup and test suggestions

src/cmd/wallet_args.rs
ardocrat: Looks like read_password is for this:
https://github.com/conradkleinespel/rpassword/blob/main/examples/read_password.rs

Cargo.toml
ardocrat: removed

libwallet/src/mwixnet/onion/crypto/dalek.rs
ardocrat: added fixed sig

libwallet/src/mwixnet/onion/crypto/dalek.rs
ardocrat: added

PR #769, opened 2026-07-08
Global config instance

ardocrat:
- Make config global instance
- Restart Tor service and write changes to Tor config on `set_tor_config` call from owner API
- Use config file path as key to store multiple configs for multiple wallets in memory


config/src/config.rs
```diff
@@ -42,6 +50,26 @@ pub const API_SECRET_FILE_NAME: &str = ".foreign_api_secret";
...
+		CONFIG_INSTANCE
+			.set(RwLock::new(config.clone()))
+			.map_err(|_e| ConfigError::Other("Global config can not be set".to_string()))?;
+	}
+	Ok(())
+}
+
+/// Get global configuration to read values.
+pub fn global_config_to_read() -> RwLockReadGuard<'static, GlobalWalletConfig> {
+	CONFIG_INSTANCE.get().unwrap().read()
```
wiesche89: Can we avoid the unwrap here? Library users can construct Owner/Foreign without going through the CLI setup path, and this would panic once a Tor config is read. Returning Option/Result would be much safer.

api/src/owner.rs
```diff
@@ -215,11 +212,16 @@ where
...
+	/// * or [`libwallet::Error`](../grin_wallet_libwallet/struct.Error.html) if an error is encountered.
+	///
 
-	pub fn set_tor_config(&self, tor_config: Option<TorConfig>) {
-		let mut lock = self.tor_config.lock();
-		*lock = tor_config;
+	pub fn set_tor_config(&self, tor_config: Option<TorConfig>) -> Result<(), Error> {
+		let mut gc = global_config_to_update();
+		gc.members.as_mut().unwrap().tor = tor_config;
+		gc.save().map_err(|e| Error::TorConfig(format!("{}", e)))?;
```
wiesche89: This changes set_tor_config from a runtime setter into a persistent rewrite of grin-wallet.toml. That can drop user comments and makes None remove [tor] permanently. Can we make the persistence explicit or document this very clearly?

impls/src/lifecycle/default.rs
```diff
@@ -176,6 +177,11 @@ where
 			config_file_name.to_str().unwrap(),
 		);
 
+		// Set global config instance.
+		set_global_config(default_config).map_err(|e| {
```
wiesche89: Should we set default_config.config_file_path before storing it globally? Otherwise a later set_tor_config can hit save() without a path, or keep using an older global config.

api/src/foreign.rs
```diff
@@ -371,7 +354,8 @@ where
 		)?;
 		match r_addr {
 			Some(a) => {
-				let tc = self.tor_config.lock();
+				let gc = global_config_to_read();
```
wiesche89: Can we scope this read lock just around the clone? Right now it can live across the Tor send, which means a waiting config update can block later readers for the duration of the network call.

config/src/config.rs
```diff
@@ -21,15 +21,23 @@ use crate::types::{
...
 use std::fs::{self, File};
 use std::io::prelude::*;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
 use toml;
 
+lazy_static! {
+	/// Global configuration instance.
+	static ref CONFIG_INSTANCE: OnceLock<RwLock<GlobalWalletConfig>> = OnceLock::new();
```
wiesche89: Small cleanup, OnceLock::new() is const, so this probably does not need lazy_static. Also, set_global_config silently ignores later configs. replacing the inner value or returning an error would be clearer.

wiesche89:
I could not find the Tor restart mentioned in the PR description. It looks like set_tor_config persists the new config, but the running listener may keep the old one until restart. Is that still planned, or should we document the behavior as is? Also worth noting the public API signature changes in the changelog.

config/src/config.rs
ardocrat: this already fixed

config/src/config.rs
ardocrat: this fixed by setting up default config if not set

api/src/owner.rs
ardocrat: comments can be saved by calling `write_to_file`

impls/src/lifecycle/default.rs
ardocrat: fixed

ardocrat:
> I could not find the Tor restart mentioned in the PR description. It looks like set_tor_config persists the new config, but the running listener may keep the old one until restart. Is that still planned, or should we document the behavior as is? Also worth noting the public API signature changes in the changelog.

this is in work

api/src/owner.rs
ardocrat: Also removed an `Option` here

wiesche89: I checked the save path again. It rewrites the whole file and recreates the standard comments, so custom user comments are still lost.

api/src/owner.rs
```diff
@@ -215,11 +212,16 @@ where
...
+	/// * `Ok(())` if the config was correctly saved
+	/// * or [`libwallet::Error`](../grin_wallet_libwallet/struct.Error.html) if an error is encountered.
+	///
 
-	pub fn set_tor_config(&self, tor_config: Option<TorConfig>) {
-		let mut lock = self.tor_config.lock();
-		*lock = tor_config;
+	pub fn set_tor_config(&self, tor_config: TorConfig) -> Result<(), Error> {
+		let mut gc = global_config_to_update();
+		gc.members.as_mut().unwrap().tor = Some(tor_config);
```
wiesche89: Could we update and save a copy first, then replace the live config only after the save succeeds? Right now a save error leaves the new values active in memory.

config/src/config.rs
wiesche89: if two wallets run in the same process, the last loaded config replaces the first one and both wallets use it. Should this remain scoped to each wallet?

config/src/config.rs
wiesche89: I checked this again and it still panics in the fallback above: when get() is None, line 69 immediately unwraps the same None. Could we use get_or_init instead?

impls/src/tor/arti.rs
```diff
@@ -96,19 +121,19 @@ pub fn start_tor_service(
 		.nickname(hs.clone())
 		.build()
 		.map_err(|e| Error::TorConfig(format!("{:?}", e)))?;
-	let running_onion_service = match client.launch_onion_service(service_config) {
+	let ((service, proxy)) = match client.launch_onion_service(service_config) {
```
wiesche89: the extra parentheses here introduce a new compiler warning. Can we remove them?

api/src/owner.rs
```diff
@@ -215,11 +212,16 @@ where
...
-	/// * Nothing
+	/// * Result Containing:
+	/// * `Ok(())` if the config was correctly saved
+	/// * or [`libwallet::Error`](../grin_wallet_libwallet/struct.Error.html) if an error is encountered.
+	///
 
-	pub fn set_tor_config(&self, tor_config: Option<TorConfig>) {
-		let mut lock = self.tor_config.lock();
-		*lock = tor_config;
+	pub fn set_tor_config(&self, tor_config: TorConfig) -> Result<(), Error> {
```
wiesche89: the comment above still describes tor_config as optional, but this parameter is no longer an Option.

impls/src/tor/arti.rs
```diff
@@ -117,56 +142,46 @@ pub fn start_tor_service(
...
-							match run_service_proxy(c.clone(), addr, request, hs.clone()).await {
+							match run_service_proxy(proxy.clone(), request, hs.clone()).await {
 								Ok(_) => {
 									info!("Tor proxy stopped");
 								}
 								Err(e) => {
 									error!("Tor proxy error: {:?}, restarting", e);
 									tokio::time::sleep(Duration::from_millis(1000)).await;
-									Box::pin(run_proxy(c, addr, request, hs)).await;
+									Box::pin(run_proxy(proxy, request, hs)).await;
```
wiesche89: Could we use a loop here instead? The recursive retry keeps another layer alive after every proxy error.

wiesche89 (commented):
I checked the latest commit again. The config is saved now, but the running Tor listener is still not restarted, and stop_tor_service is never called. Is that part still missing?

api/src/owner.rs
ardocrat: probably old commit, currently it saves comments as I tested

api/src/owner.rs
ardocrat: we can try to save file, only then update live then

impls/src/tor/arti.rs
ardocrat: fixed

impls/src/tor/arti.rs
ardocrat: added loop here

api/src/owner.rs
ardocrat: fixed

config/src/config.rs
ardocrat: added

config/src/config.rs
ardocrat: at `grin-wallet` we use single wallet folder, at GUI I use name of folder as id, think this case will not work for CLI, so we can provide optional id on config creation to help with integration

ardocrat: added ability to store multiple configurations from config path as key

src/cmd/wallet_args.rs
```diff
@@ -1175,13 +1174,11 @@ where
 		("listen", Some(args)) => {
 			let mut c = wallet_config.clone();
 			let mut t = tor_config.clone();
-			let a = arg_parse!(parse_listen_args(&mut c, &mut t, &args));
+			arg_parse!(parse_listen_args(&mut c, &mut t, &args));
```
wiesche89: It looks like --no_tor and --bridge are written to t, but t is not passed to the listener anymore. So both options have no effect now. Can we pass the updated Tor config to the listener again?

impls/src/lifecycle/default.rs
```diff
@@ -142,6 +144,8 @@ where
 
 		// If config exists but the datadir return ok
 		if config_file_name.exists() {
+			// Set global config instance.
+			set_global_config(default_config);
```
wiesche89: The config file already exists here, but we put default_config into the global state. Should we load the existing file instead? Otherwise the in memory config can be different from the file.

config/src/config.rs
```diff
@@ -42,6 +60,106 @@ pub const API_SECRET_FILE_NAME: &str = ".foreign_api_secret";
...
+		}
+		Some(path) => {
+			let path = path.to_str().unwrap();
+			{
+				let configs = CONFIG_INSTANCES.read();
+				if let Some(config) = configs.get(path) {
+					return config.clone();
+				}
+			}
+			let mut default_config = GlobalWalletConfig::default();
```
wiesche89: I think a missing cache entry should not create and store a default config here. set_tor_config may later write these defaults over an existing config file. There is also a small race between releasing the read lock and taking the write lock. Could we load the file or return an error instead?

impls/src/tor/arti.rs
```diff
@@ -183,6 +189,8 @@ where
 	}
 	.to_string();
 	let timeout = tor_config.request_timeout();
+	let (state_path, cache_path) = state_cache_paths(tor_config.send_config_dir.as_str());
```
wiesche89: start_tor_client uses send_config_dir/tor/sender, but here we use send_config_dir directly. Because of this, the cached client does not match and Arti creates another client in a different directory. Can we use the same path in both places?

controller/src/command.rs
```diff
@@ -358,76 +341,74 @@ where
...
+			}
+		};
+	}
 
 	if args.estimate_selection_strategies {
 		return Ok(());
 	}
 
-	let tor_config = match tor_config {
+	let tc = {
```
wiesche89: send now reads the Tor config from the global state, while receive and process_invoice still use the resolved config with the legacy fallback. With an old config without a [tor] section, send may not use Tor anymore. Can we use the same config source for all three commands?

controller/src/controller.rs
```diff
@@ -265,99 +251,121 @@ where
...
-			Ok(service) => {
-				warn!(
-					"Starting Tor Hidden Service for API listener at address {}, binding to {}",
-					onion_address, addr
-				);
-				Ok(Some(service))
+
+		// Start thread to check if restart is needed.
+		let (restart_tx, restart_rx) = std::sync::mpsc::channel::<()>();
+		add_global_config_listener(&config_path, "foreign_listener", restart_tx);
```
wiesche89: The config listener is registered only after the HTTP and Tor services were started. If the config changes during the Tor startup, the notification is lost and the service keeps the old settings. Can we register the listener before starting the services?

config/src/config.rs
```diff
@@ -476,4 +590,23 @@ impl GlobalWalletConfig {
 			.replace("WARN", "Warning")
 			.replace("ERROR", "Error")
 	}
+
+	/// Save config to file and update global state after editing.
+	pub fn save(&mut self) -> Result<(), ConfigError> {
+		if let Some(path) = self.config_file_path.clone() {
+			let res = self.write_to_file(path.to_str().unwrap(), false, None, None);
```
wiesche89: This can now be called while the wallet is running, but write_to_file writes directly to the real config file. A crash or write error could leave a broken config behind. Would it be safer to write to a temporary file and rename it afterwards?

config/src/config.rs
```diff
@@ -21,17 +21,35 @@ use crate::types::{
...
 
+lazy_static! {
+	/// Global configuration instance.
+	static ref CONFIG_INSTANCE: OnceLock<RwLock<GlobalWalletConfig>> = OnceLock::new();
+	/// Global configuration instances mapped to config path.
+	static ref CONFIG_INSTANCES: Arc<RwLock<HashMap<String, GlobalWalletConfig>>> = Arc::new(RwLock::new(HashMap::new()));
+	/// Global configuration change listener.
+	static ref CONFIG_CHANGE_LISTENER: Arc<RwLock<HashMap<String, Sender<()>>>> = Arc::new(RwLock::new(HashMap::new()));
+	/// Global configuration change listeners mapped to config path.
+	static ref CONFIG_CHANGE_LISTENERS: Arc<RwLock<HashMap<String, HashMap<String, Sender<()>>>>> = Arc::new(RwLock::new(HashMap::new()));
```
wiesche89: Do we need separate global maps for configs and listeners, plus the additional None variants? Maybe one PathBuf based registry containing both config and listeners would make this simpler and also avoid the repeated to_str().unwrap() calls.

config/src/types.rs
```diff
@@ -141,6 +141,9 @@ pub enum ConfigError {
 
 	/// Path doesn't exist
 	PathNotFoundError(String),
+
+	/// Other errors
+	Other(String),
```
wiesche89: I could not find a use of this new error variant in the PR. Can we remove it for now and add it when it is actually needed?

impls/src/tor/types.rs
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+
+/// Arti client runtime wrapper.
+#[derive(Clone)]
+pub struct ArtiRuntimeWrapper {
+	pub runtime: tor_rtcompat::tokio::TokioNativeTlsRuntime,
+}
+
+impl ArtiRuntimeWrapper {
+	pub fn create() -> Result<ArtiRuntimeWrapper, std::io::Error> {
```
wiesche89: Does this wrapper need its own public module and a public constructor? It still has only one internal user, so maybe it can stay private near the Arti implementation.

controller/src/controller.rs
```diff
@@ -265,99 +251,121 @@ where
...
+			.start(socket_addr, router, tls_config.clone(), api_chan)
+			.map_err(|_| Error::GenericError("API thread failed to start".to_string()))?;
+		warn!("HTTP Foreign listener started.");
+
+		// Need to keep external process in scope while the listener is running.
+		let _tor_service = if tor_config.use_tor_listener {
+			let use_integrated = tor_config.use_integrated.unwrap_or(false);
+
+			let res: Result<Option<tor_process::TorProcess>, Error> = if use_integrated {
+				start_tor_service(sec_key, &tor_dir, addr, &tor_config.clone())?;
```
wiesche89: Both Tor start calls use ? here, so the Err branch below can never handle a startup error. Before this change the local HTTP listener stayed available when Tor failed, but now foreign_listener returns immediately. Could we keep the error inside res and preserve the fallback?

api/src/owner_rpc.rs
```diff
@@ -1924,7 +1925,7 @@ pub trait OwnerRpc {
 	# , 0, false, false, false, false);
 	```
 	*/
-	fn set_tor_config(&self, tor_config: Option<TorConfig>) -> Result<(), Error>;
+	fn set_tor_config(&self, tor_config: TorConfig) -> Result<(), Error>;
```
wiesche89: This changes the public RPC parameter from Option<TorConfig> to TorConfig. Existing clients sending null will now fail, and there is no way to clear the Tor config anymore. Do we need this breaking API change, or can we keep Option and persist None as well?

controller/src/controller.rs
```diff
@@ -265,99 +251,121 @@ where
...
-			let sp_address = SlatepackAddress::try_from(onion_address.clone())?;
-			let qr_string = match QrCode::new(sp_address.to_string()) {
-				Ok(qr) => qr.to_string(false, 3),
-				Err(_) => "Failed to generate QR code!".to_string(),
-			};
-			warn!("Slatepack Address is: {}\n{}", sp_address, qr_string);
-		} else {
-			warn!("Listener is available on {}", addr);
-		}
+		return res;
```
wiesche89, line 386: On a normal API shutdown, the integrated Tor service is still kept in the global service map. It is removed only during a config restart, so the service and proxy can remain alive after the HTTP listener has stopped. Could we clean it up before returning here as well?

wiesche89 (commented):
Can we add tests for --no_tor and --bridge, config changes during Tor startup, loading an existing config and reusing the Arti client? These are important new paths introduced by this change.

api/src/owner.rs
wiesche89: I reproduced this again with the current PR head. Saving the Tor config rewrites the file and preserves the generated standard comments, but custom user comments are still removed.

controller/src/command.rs
wiesche89: The same raw global config lookup is also used by Owner::init_send_tx, Owner::process_invoice_tx and Foreign::receive_tx. With a legacy config without a [tor] section, these API paths silently skip Tor as well. Could we centralize the legacy fallback in one config accessor?

api/src/owner_rpc.rs
ardocrat: Optional tor config conflicts with logic when we load default torconfig on send, so Option is no longer needed, to disable send users now shoud set `skip_send_attempt` (not used before) and `use_tor_listener`

controller/src/command.rs
ardocrat: Need to decide, should we always load tor config or not, such construction called before send:
image

src/cmd/wallet_args.rs
ardocrat: passing initial tor config with args, will be overwritten with new config on restart

controller/src/controller.rs
ardocrat: added service stop

controller/src/controller.rs
ardocrat: added error handling

impls/src/tor/types.rs
ardocrat: moved to arti.rs

impls/src/lifecycle/default.rs
ardocrat: should we use provided values to write to config here or not?

ardocrat: We also can check existing config in the beginning of this function

config/src/config.rs
ardocrat: added an error on config getting

impls/src/tor/arti.rs
ardocrat: for arti it does not matter which dir to use, I will use `send_config_dir` for both send and listener adding `arti` subpath

controller/src/controller.rs
ardocrat: moved registration on top

config/src/config.rs
ardocrat: added saving to .tmp, then rename to original

config/src/types.rs
ardocrat: removed

config/src/config.rs
ardocrat: then we also need to make `config_file_path` at GlobalWalletConfig mandatory

impls/src/lifecycle/default.rs
ardocrat: Added writing to existing file

config/src/config.rs
ardocrat: Made config_file_path not optional, just like configmembers

api/src/owner_rpc.rs
wiesche89: I agree that the config can be non optional internally. My concern is the RPC boundary. Both master and staging currently accept null. Could we keep Option there for compatibility and translate None into a resolved config with sending and listening disabled?

controller/src/command.rs
wiesche89: I would always resolve the Tor config once when loading the wallet config. We can keep the stored field optional for old files, but all CLI and API paths should receive the same effective config. The fallback also needs to use the wallet’s data_file_dir instead of the default .

impls/src/lifecycle/default.rs
wiesche89: I would check and load the existing config at the beginning. Since this is create_config, I would not overwrite an existing file with the provided creation values. The existing config should stay authoritative, while initializing an already existing wallet should return an error.

config/src/config.rs
wiesche89: Using PathBuf as the registry key makes sense. I would only make the path mandatory for registration and persistence, unless we are sure that library users no longer need in memory configs without a file path.

controller/src/controller.rs
```diff
@@ -265,99 +252,133 @@ where
...
-				Ok(qr) => qr.to_string(false, 3),
-				Err(_) => "Failed to generate QR code!".to_string(),
-			};
-			warn!("Slatepack Address is: {}\n{}", sp_address, qr_string);
-		} else {
-			warn!("Listener is available on {}", addr);
+
+		if restart_needed.load(Ordering::Relaxed) {
+			let config = get_global_config(&config_path)?;
+			tor_config = config.members.unwrap().tor.unwrap_or(TorConfig::default());
```
wiesche89: The initial listener uses the --no_tor and --bridge overrides, but after a config-triggered restart this replaces them with the persisted Tor config. Could we apply the CLI overrides again after every reload so they remain valid for the lifetime of the command?

controller/src/controller.rs
```diff
@@ -265,99 +252,133 @@ where
...
-	let api_handler_v2 = ForeignAPIHandlerV2::new(
-		wallet,
-		keychain_mask,
-		test_mode,
-		Mutex::new(Some(tor_config.clone())),
-	);
-	let mut router = Router::new();
+	loop {
+		let (restart_tx, restart_rx) = std::sync::mpsc::channel::<()>();
+		add_global_config_listener(&config_path, "foreign_listener", restart_tx);
```
wiesche89: tor_config was already resolved before entering this function. A config change between that read and registering the listener here can still be missed. Could we reload the current config immediately after the listener has been registered?

config/src/config.rs
```diff
@@ -42,6 +60,105 @@ pub const API_SECRET_FILE_NAME: &str = ".foreign_api_secret";
...
+			let path = path.to_str().unwrap();
+			{
+				let configs = CONFIG_INSTANCES.read();
+				if let Some(config) = configs.get(path) {
+					return Ok(config.clone());
+				}
+			}
+			let config = GlobalWalletConfig::new(path)?;
+			let mut configs = CONFIG_INSTANCES.write();
+			configs.insert(path.to_string(), config.clone());
```
wiesche89: While the config file is read without holding the map lock, another thread can save and cache a newer config. This insert can then overwrite it with the older snapshot. Could we check the cache again after taking the write lock before inserting?

api/src/owner.rs
```diff
@@ -198,28 +200,33 @@ where
...
```
wiesche89: Owner::new still allows config_path = None, but get_global_config(None) creates a default Tor config while this setter always fails because save() has no path. Could we either support a real in-memory setter here or require a config path for the persistent operation?

config/src/config.rs
```diff
@@ -476,4 +589,24 @@ impl GlobalWalletConfig {
 			.replace("WARN", "Warning")
 			.replace("ERROR", "Error")
 	}
+
+	/// Save config to file and update global state after editing.
+	pub fn save(&mut self) -> Result<(), ConfigError> {
+		if let Some(path) = self.config_file_path.clone() {
+			let tmp_path = format!("{}.tmp", path.to_str().unwrap());
+			let res = self.write_to_file(tmp_path.as_str(), false, None, None);
```
wiesche89: save() recreates the config and only adds the standard comments again, so custom comments are lost. It also uses the same temporary file for every save and may change the original file permissions. Could we preserve the existing comments and permissions and use a unique temporary file?

controller/src/controller.rs
```diff
@@ -265,99 +252,133 @@ where
...
-			Err(e) => {
-				warn!("Unable to start TOR listener");
-				error!("Tor Error: {}", e);
-				warn!("Listener is available on {}", addr);
-				Err(e)
+
+		// Start thread to check if restart is needed.
+		let restart_needed = Arc::new(AtomicBool::new(false));
+		let t_restart_needed = restart_needed.clone();
+		thread::spawn(move || {
```
wiesche89, line 365: On a normal api shutdown this sender remains in the global registry, so the worker keeps blocking on recv() and cannot exit. Could we unregister the listener, drop its sender and join the worker before returning?

impls/src/tor/arti.rs
```diff
@@ -42,19 +42,39 @@ use tor_hsrproxy::config::{
...
+	static ref ARTI_CLIENT_CONFIG: LazyLock<Mutex<Option<(Arc<TorClient<TokioNativeTlsRuntime>>, TorClientConfig)>>> =
 		LazyLock::new(|| Mutex::new(None));
+	/// Running services, where key is onion address.
+	static ref ARTI_PROXY_SERVICES: LazyLock<Mutex<HashMap<String, (Arc<RunningOnionService>, Arc<OnionServiceReverseProxy>)>>> =
+		LazyLock::new(|| Mutex::new(HashMap::new()));
+}
+
+/// Arti client runtime wrapper.
+#[derive(Clone)]
+pub struct ArtiRuntimeWrapper {
```
wiesche89: Moving the wrapper next to the arti implementation is cleaner, but the struct, field and constructor are still public although they only have internal users. Could they stay private to this module?

api/src/owner.rs
```diff
@@ -198,28 +200,33 @@ where
...
 			shared_key: Arc::new(Mutex::new(None)),
 			updater,
 			updater_running,
 			status_tx: Mutex::new(Some(tx)),
 			updater_messages,
-			tor_config: Mutex::new(None),
 		}
 	}
 
 	/// Set the TOR configuration for this instance of the OwnerAPI, used during
```
wiesche89, line 232: The documentation still describes this as setting Tor config for this OwnerAPI instance. It now persists the complete wallet config and notifies every listener registered for the same path. Could we update the public documentation to describe these side effects?

tests/common/mod.rs
```diff
@@ -191,7 +192,9 @@ pub fn initial_setup_wallet(dir_name: &str, wallet_name: &str) -> GlobalWalletCo
 	let _ = fs::create_dir_all(current_dir.clone());
 	let mut config_file_name = current_dir.clone();
 	config_file_name.push("grin-wallet.toml");
-	GlobalWalletConfig::new(config_file_name.to_str().unwrap()).unwrap()
+	let config = GlobalWalletConfig::new(config_file_name.to_str().unwrap()).unwrap();
+	set_global_config(config.clone());
```
wiesche89: The existing tests are updated to use the global config, but I could not find tests for its new behavior. Could we cover legacy and pathless configs, safe concurrent saves, CLI overrides after restart, and listener cleanup?

impls/src/lifecycle/default.rs
ardocrat: returning an error if wallet config file exists now

controller/src/command.rs
ardocrat: providing default tor config also to send if not set

api/src/owner_rpc.rs
ardocrat: setting `skip_send_attempt` and `use_tor_listener` to null if none was provided

controller/src/controller.rs
ardocrat: passing tor config from argument here

ardocrat: > The initial listener uses the --no_tor and --bridge overrides, but after a config-triggered restart this replaces them with the persisted Tor config. Could we apply the CLI overrides again after every reload so they remain valid for the lifetime of the command?

yes, this is expected behaviour if user changed tor config and wants to use new values

controller/src/controller.rs
ardocrat: passed tor config can contain args from CLI, so they can be overriden by tor_config from global configuration

config/src/config.rs
ardocrat: added a check for key

api/src/owner.rs
ardocrat: changed this operation by disabling sending or launch listener

impls/src/tor/arti.rs
ardocrat: made private

api/src/owner.rs
ardocrat: updated

tests/common/mod.rs
ardocrat: Made configs non-pathless already, CLI args will not be update after foreign listener restart as expected behaviour on config change, listener cleanup can be checked

controller/src/controller.rs
ardocrat: added removal of config change listener

config/src/config.rs
ardocrat: passing migration `true` to migrate comments

config/src/config.rs
ardocrat: this was fixed at next commit, now we are sure path to config file can be always provided

config/src/config.rs
wiesche89: save() still uses one shared .tmp file, may change 0600 to 0644, and loses some custom comments. Could we use a unique temporary file, preserve permissions, and pass the actual config version to migrate_comments?

config/src/config.rs
```diff
@@ -42,6 +54,76 @@ pub const API_SECRET_FILE_NAME: &str = ".foreign_api_secret";
...
+		if let Some((config, _)) = configs.get(config_path) {
+			return Ok(config.clone());
+		}
+	}
+	let config = GlobalWalletConfig::new(config_path.to_str().unwrap())?;
+	let mut configs = CONFIG_INSTANCES.write();
+	if !configs.contains_key(config_path) {
+		configs.insert(config_path.clone(), (config.clone(), HashMap::new()));
+	}
+	Ok(config)
```
wiesche89: The second cache check prevents overwriting a newer config, but this still returns the older snapshot read from disk. Could we return the value currently stored in the cache?

api/src/owner.rs
```diff
@@ -198,28 +201,41 @@ where
...
-		*lock = tor_config;
+	pub fn set_tor_config(&self, tor_config: Option<TorConfig>) -> Result<(), Error> {
+		let mut gc = get_global_config(&self.config_path)?;
+		if let Some(tor_config) = tor_config {
+			gc.members.tor = Some(tor_config);
+		} else if let Some(tor) = gc.members.tor.as_mut() {
+			tor.use_tor_listener = false;
+			tor.skip_send_attempt = Some(true);
+		} else {
+			return Ok(());
```
wiesche89: If [tor] is missing, set_tor_config(null) is a no op and does not restart the listener. Could we create and save a disabled TorConfig instead?

controller/src/controller.rs
wiesche89: Could we reload the config after registering the listener and then reapply --no_tor and --bridge? This closes the race while keeping explicit CLI overrides active for the process lifetime.

config/src/config.rs
```diff
@@ -42,6 +54,76 @@ pub const API_SECRET_FILE_NAME: &str = ".foreign_api_secret";
...
+
+/// Get global configuration using provided path.
+pub fn get_global_config(config_path: &PathBuf) -> Result<GlobalWalletConfig, ConfigError> {
+	{
+		let configs = CONFIG_INSTANCES.read();
+		if let Some((config, _)) = configs.get(config_path) {
+			return Ok(config.clone());
+		}
+	}
+	let config = GlobalWalletConfig::new(config_path.to_str().unwrap())?;
```
wiesche89: This to_str().unwrap() can panic for a valid non UTF-8 path. Could GlobalWalletConfig::new accept a &Path directly?

config/src/config.rs
ardocrat: added return of config from cache if exists

api/src/owner.rs
ardocrat: Added this

controller/src/controller.rs
ardocrat: added

config/src/config.rs
ardocrat: added

config/src/config.rs
ardocrat: Added unique file name, setting permissions and providing config version.

tests/common/mod.rs
wiesche89: The new tests cover legacy Tor config resolution, concurrent saves and cache consistency, comment and permission preservation, null disabling Tor, listener reloads with CLI overrides, listener cleanup, and config errors before wallet state is changed.

wiesche89 (approved):

Issue #770, opened 2026-07-09
No mainnet/testnet check in send/receive/finalize


noobvie:
**Describe the bug**
There is no check when sending grin to different network, I tested to send grin from mainnet to testnet and send tgrin from testnet to mainnet, confirmed coin lost. I asked Claude to write full scenerio I tested and the suggestions. Hope someone can re-produce the tests I did. You may get few coins from drop.grin.money if you need to test.

## Summary

Slatepack addresses already encode which network they belong to (per RFC-0015: HRP `grin`
for mainnet, `tgrin` for testnet/floonet — e.g. `grin1...` vs `tgrin1...`). Despite this,
nothing in the `send` / `receive` / `finalize` pipeline checks the destination address's
HRP against the wallet's own `chain_type`. Completing a full manual Slatepack round-trip
(no Tor, no live network connection — just copy/pasting armored text between two wallets)
between a mainnet wallet and a testnet wallet results in a **confirmed on-chain spend on
the sender's own chain**, while the receiving wallet's local transaction ledger never
reflects it (it perpetually shows "Sent, awaiting finalization" because it only scans its
own network's chain, which never sees the transaction). The result is permanent, silent
fund loss for the sender, with no error anywhere in the process.

Verified end-to-end in **both directions**.

## Environment

- `grin-wallet` version: v5.4.1
- Two wallets: one initialized/run with default (mainnet) config, one with `--testnet`
- No Tor required — pure manual/offline Slatepack exchange (`send -d`, paste, `receive`,
  paste back, `finalize`)

## Steps to reproduce

**Case A — mainnet sender → tgrin (testnet) destination:**
1. Mainnet wallet: `grin-wallet send -d tgrin1xxx... 100` toward a real Slatepack address
   belonging to a wallet running `--testnet`. Produces armored Slatepack message S1. No
   warning about the destination being on a different network.
2. Copy S1 to the testnet wallet → `grin-wallet receive` (paste S1) → produces S2.
3. Copy S2 back to the mainnet wallet → `grin-wallet finalize` (paste S2).
4. Mainnet wallet posts the final tx to its own (mainnet) node.
   **Result:** `grin-wallet txs` on the mainnet (sending) wallet shows the transaction as
   **Confirmed** — the mainnet coin is spent for real. On the testnet (receiving) wallet,
   the same transaction sits indefinitely as **"Sent, Awaiting Finalization"** and never
   moves, because that wallet only scans the testnet chain and the transaction was actually
[...]

PR #773, opened 2026-07-21
Wrong Slatepack address fix

ardocrat:
- check network prefix on Slatepack parsing
- do not create tx when wrong Slatepack address specified
- fix tests with bad addresses

Closes https://github.com/mimblewimble/grin-wallet/issues/770

### DIFF
--- a/api/src/foreign_rpc.rs
+++ b/api/src/foreign_rpc.rs
@@ -21,6 +21,7 @@ use crate::libwallet::{
 };
 use crate::{Foreign, ForeignCheckMiddlewareFn};
 use easy_jsonrpc_mw;
+use libwallet::SlatepackAddress;
 
 /// Public definition used to generate Foreign jsonrpc api.
 /// * When running `grin-wallet listen` with defaults, the V2 api is available at
@@ -309,6 +310,10 @@ where
 	) -> Result<VersionedSlate, Error> {
 		let version = in_slate.version();
 		let slate_from = Slate::from(in_slate);
+		let dest = match dest {
+			None => None,
+			Some(a) => SlatepackAddress::try_from(a.as_str()).ok(),
+		};
 		let out_slate = Foreign::receive_tx(
 			self,
 			&slate_from,
--- a/api/src/foreign_rpc.rs
+++ b/api/src/foreign_rpc.rs
@@ -21,7 +21,9 @@ use crate::libwallet::{
 };
 use crate::{Foreign, ForeignCheckMiddlewareFn};
 use easy_jsonrpc_mw;
+use grin_wallet_config::initial_setup_wallet;
 use libwallet::SlatepackAddress;
+use std::path::PathBuf;
 
 /// Public definition used to generate Foreign jsonrpc api.
 /// * When running `grin-wallet listen` with defaults, the V2 api is available at
@@ -375,6 +377,14 @@ pub fn run_doctest_foreign(
 	let _ = fs::remove_dir_all(test_dir);
 	global::set_local_chain_type(ChainTypes::AutomatedTesting);
 
+	let _ = fs::create_dir_all(test_dir);
+	let config = initial_setup_wallet(
+		&ChainTypes::AutomatedTesting,
+		Some(PathBuf::from(test_dir)),
+		false,
+	)
+	.unwrap();
+
 	let mut wallet_proxy: WalletProxy<
 		DefaultLCProvider<LocalWalletClient, ExtKeychain>,
 		LocalWalletClient,
@@ -523,8 +533,20 @@ pub fn run_doctest_foreign(
 	}
 
 	let mut api_foreign = match init_invoice_tx {
-		false => Foreign::new(wallet1, mask1, Some(test_check_middleware), true),
-		true => Foreign::new(wallet2, mask2, Some(test_check_middleware), true),
+		false => Foreign::new(
+			wallet1,
+			config.config_file_path,
+			mask1,
+			Some(test_check_middleware),
+			true,
+		),
+		true => Foreign::new(
+			wallet2,
+			config.config_file_path,
+			mask2,
+			Some(test_check_middleware),
+			true,
+		),
 	};
 	api_foreign.doctest_mode = true;
 	let foreign_api = &api_foreign as &dyn ForeignRpc;
--- a/api/src/foreign_rpc.rs
+++ b/api/src/foreign_rpc.rs
@@ -312,7 +312,13 @@ where
 		let slate_from = Slate::from(in_slate);
 		let dest = match dest {
 			None => None,
-			Some(a) => SlatepackAddress::try_from(a.as_str()).ok(),
+			Some(a) => match SlatepackAddress::try_from(a.as_str()) {
+				Ok(d) => Some(d),
+				Err(_) => {
+					error!("Error parsing Slatepack address: {}", a);
+					None
+				}
+			},
 		};
 		let out_slate = Foreign::receive_tx(
 			self,
--- a/api/src/owner.rs
+++ b/api/src/owner.rs
@@ -38,6 +38,7 @@ use crate::util::secp::{key::SecretKey, pedersen::Commitment};
 use crate::util::{from_hex, static_secp_instance, Mutex, ZeroingString};
 use grin_wallet_util::OnionV3Address;
 use libwallet::api_impl::types::update_tx_slate_state;
+use libwallet::InitTxSendArgs;
 use std::convert::TryFrom;
 use std::fs::File;
 use std::io::Write;
@@ -665,11 +666,24 @@ where
 		args: InitTxArgs,
 	) -> Result<Slate, Error> {
 		let send_args = args.send_args.clone();
+		let dest = match send_args.as_ref() {
+			None => None,
+			Some(args) => {
+				if args.dest.is_empty() {
+					None
+				} else {
+					let a = SlatepackAddress::try_from(args.dest.as_str())?;
+					Some(a)
+				}
+			}
+		};
+
 		let slate = {
 			let mut w_lock = self.wallet_inst.lock();
 			let w = w_lock.lc_provider()?.wallet_inst()?;
 			owner::init_send_tx(w, keychain_mask, args, self.doctest_mode)?
 		};
+
 		// Helper functionality. If send arguments exist, attempt to send sync and
 		// finalize
 		match send_args {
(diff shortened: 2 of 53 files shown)


libwallet/src/slatepack/address.rs
```diff
@@ -86,6 +86,16 @@ impl Display for SlatepackAddress {
 impl TryFrom<&str> for SlatepackAddress {
 	type Error = Error;
 	fn try_from(encoded: &str) -> Result<Self, Self::Error> {
+		let prefix = match global::get_chain_type() {
+			global::ChainTypes::Mainnet => "grin",
+			_ => "tgrin",
+		};
+		if !encoded.starts_with(prefix) {
```
wiesche89: I think we should compare the decoded human-readable part here. With starts_with, an address using grinextra is accepted on mainnet, while a valid uppercase GRIN1... address is rejected. Could we decode first, compare the human-readable part exactly, and add tests for both network directions and casing?

api/src/owner.rs
```diff
@@ -2502,7 +2504,7 @@ where
 /// attempt to send slate synchronously with TOR
 pub fn try_slatepack_sync_workflow(
 	slate: &Slate,
-	dest: &str,
+	dest: &SlatepackAddress,
```
wiesche89, line 2580: A SlatepackAddress does not guarantee the correct network because human-readable part and pub_key are public. A Rust caller can construct a tgrin address while mainnet is active, and this path no longer checks the human-readable part. Could we validate the network before init_send_tx or process_invoice_tx creates any wallet state?

libwallet/src/api_impl/types.rs
```diff
@@ -98,7 +98,7 @@ pub struct InitTxArgs {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct InitTxSendArgs {
 	/// The destination, contents will depend on the particular method
-	pub dest: String,
+	pub dest: Option<SlatepackAddress>,
```
wiesche89: send_args represents the combined init-and-send path, but making dest optional now accepts a missing or null destination. The wallet transaction is created first and then returned without any send attempt. Could this remain a required SlatepackAddress, with Option only used by the separate CLI arguments?

api/src/foreign_rpc.rs
```diff
@@ -309,6 +310,10 @@ where
 	) -> Result<VersionedSlate, Error> {
 		let version = in_slate.version();
 		let slate_from = Slate::from(in_slate);
+		let dest = match dest {
+			None => None,
+			Some(a) => SlatepackAddress::try_from(a.as_str()).ok(),
```
wiesche89: An invalid or wrong network callback address is silently converted to None here. Keeping receive_tx successful matches the previous best-effort behavior, but the failed callback was at least logged before. Could we log the parse error so the missing Tor response is not invisible?

src/cmd/wallet_args.rs
```diff
@@ -727,6 +726,26 @@ pub fn parse_issue_invoice_args(
 	})
 }
 
+fn get_slatepack_address_arg(
+	name: &str,
+	args: &ArgMatches,
+) -> Result<Option<SlatepackAddress>, ParseError> {
+	if args.is_present(name) {
+		if let Some(dest) = args.value_of(name) {
+			match SlatepackAddress::try_from(dest) {
```
wiesche89: The help for send -d still advertises HTTP listener addresses as deprecated, but this parser now rejects every destination that is not a Slatepack address. Could we remove the HTTP note, or preserve the legacy path if it is still supported?

libwallet/src/slatepack/address.rs
ardocrat: we can using `to_lowercase()` to check valid prefix registry

ardocrat: added tests

src/cmd/wallet_args.rs
ardocrat: removed comment

api/src/owner.rs
ardocrat: > Could we validate the network before init_send_tx or process_invoice_tx creates any wallet state?

`InitTxSendArgs` and `SendArgs` are using to validate address at CLI, for external callers we can keep strings inside and parsing when taking them out

api/src/foreign_rpc.rs
ardocrat: added logging

libwallet/src/api_impl/types.rs
wiesche89: Creating a slate without sending is valid, but send_args: None already covers this. Allowing send_args with dest: None may hide malformed requests. If intentional, it should be documented and tested.

api/src/owner.rs
wiesche89: Keeping strings for external callers can work, but they should be parsed at the start of init_send_tx and process_invoice_tx, before the internal owner call creates wallet state. Currently, a Rust caller can still construct a SlatepackAddress with the wrong human-readable part, and this path never validates it. Could we validate it before creating any transaction state and cover this with a regression test?

libwallet/src/slatepack/address.rs
```diff
@@ -86,6 +86,16 @@ impl Display for SlatepackAddress {
 impl TryFrom<&str> for SlatepackAddress {
 	type Error = Error;
 	fn try_from(encoded: &str) -> Result<Self, Self::Error> {
+		let prefix = match global::get_chain_type() {
+			global::ChainTypes::Mainnet => "grin",
+			_ => "tgrin",
+		};
+		if !encoded.to_lowercase().starts_with(prefix) {
```
wiesche89: Lowercasing fixes uppercase addresses, but starts_with still accepts a valid Bech32 address using tgrinextra or grinextra. I reproduced this at the current head. Could we decode first, compare the human-readable part exactly, and test both chain directions plus an extended prefix?

src/cmd/wallet_args.rs
```diff
@@ -483,7 +483,7 @@ pub fn parse_send_args(args: &ArgMatches) -> Result<command::SendArgs, ParseErro
 	let late_lock = args.is_present("late_lock");
 
 	// dest
-	let dest = args.value_of("dest").unwrap_or_else(|| "default");
+	let dest = get_slatepack_address_arg("dest", args)?;
```
wiesche89: --estimate-selection does not use the destination, but the address is now parsed first and can abort an estimate-only request. Could we skip destination validation for estimates, or only parse it for an actual send?

api/src/owner.rs
ardocrat: added address parsing inside owner methods

libwallet/src/api_impl/types.rs
ardocrat: OK, the only place where we are setting `send_args` currently is Owner API with calls `init_send_tx` or `process_invoice_tx`, will make String here and will parse when needed

src/cmd/wallet_args.rs
ardocrat: providing None if dest was not present now

libwallet/src/slatepack/address.rs
ardocrat: I tried to use some wrong addresses and got error on checksum and length, whats actually issue is here and can you provide wrong slatepack address to test?

wiesche89: Simply changing the prefix breaks the checksum. 
This correctly encoded address is still accepted on testnet, 
although its prefix is tgrinextra instead of tgrin:

tgrinextra1xtxavwfgs48ckf3gk8wwgcndmn0nt4tvkl8a7ltyejjcy2mc6nfs4ylecn

I tested it on the current version. Could we require the prefix to match exactly?

This address is definitely invalid, but it currently returns Ok(...). It should return Err(...) because the prefix is tgrinextra instead of exactly tgrin.

libwallet/src/api_impl/types.rs
```diff
@@ -78,7 +78,7 @@ pub struct InitTxArgs {
 	pub ttl_blocks: Option<u64>,
 	/// If set, require a payment proof for the particular recipient
 	#[serde(default)]
-	pub payment_proof_recipient_address: Option<SlatepackAddress>,
+	pub payment_proof_recipient_address: Option<String>,
```
wiesche89: I see why the raw string is parsed at the API boundary. However, changing this public field from Option<SlatepackAddress> to Option<String> breaks existing Rust callers. Could we keep the existing type and validate its network before the wallet call?

controller/src/command.rs
```diff
@@ -355,6 +355,13 @@ where
 	C: NodeClient + 'static,
 	K: keychain::Keychain + 'static,
 {
+	let dest = if let Some(d) = args.dest {
+		let a = SlatepackAddress::try_from(d.as_str())?;
```
wiesche89: The destination is still parsed before the estimate-only branch. This means an invalid -d stops --estimate-selection, although the destination is not used there. Could we parse it only after the estimate has returned?

api/src/owner.rs
```diff
@@ -665,11 +666,24 @@ where
 		args: InitTxArgs,
 	) -> Result<Slate, Error> {
 		let send_args = args.send_args.clone();
+		let dest = match send_args.as_ref() {
+			None => None,
+			Some(args) => {
+				if args.dest.is_empty() {
+					None
+				} else {
+					let a = SlatepackAddress::try_from(args.dest.as_str())?;
```
wiesche89: Parsing the address before the wallet call looks correct. Could we add a regression test confirming that a wrong-network destination returns an error before any transaction context or tx-log entry is created?

controller/src/command.rs
ardocrat: we can ignore dest parsing at all for estimate

libwallet/src/api_impl/types.rs
ardocrat: bring back SlatepackAddress, added validation at owner api

libwallet/src/slatepack/address.rs
ardocrat: getting prefix from bech32::decode

api/src/owner.rs
ardocrat: added tests

libwallet/src/slatepack/address.rs
```diff
@@ -75,6 +75,13 @@ impl SlatepackAddress {
 		let x_key = xDalekPublicKey::try_from(self)?;
 		Ok(bech32::encode("age", x_key.as_bytes().to_base32())?.to_string())
 	}
+
+	/// Check if address network is valid.
+	pub fn valid_network(&self) -> Result<bool, Error> {
```
wiesche89: valid_network() can panic when a manually constructed address contains an invalid human readable part, because converting it to a string uses unwrap(). Could we compare self.hrp directly and add a small test for an invalid value?

ardocrat: fixed and added test

PR #778, opened 2026-07-26
RFC: Explore running a node and wallet in the same process

wiesche89:
Adds an optional `--embedded-node` mode that runs a Grin node and wallet in the same process.

Example:
grin-wallet --embedded-node --testnet owner_api --run_foreign


ardocrat (commented):
Does it mean wallet should be able to use `client` commands to see node stats?

aglkm:
What's the use-case for this feature?

wiesche89:
> Does it mean wallet should be able to use `client` commands to see node stats?

Not at the moment. This PR only starts the node and connects the wallet to its local API. It doesn’t expose the grin client commands through grin-wallet


> What's the use-case for this feature?

This is mainly meant as a starting point for discussion.. The idea is to make setups like a Stratum server with a wallet, or a wallet bot, easier to run. They can start the wallet and node together instead of having to manage a separate grin process. It’s optional, so the normal wallet setup stays unchanged.

PR #779, opened 2026-07-27
Use native tls for client

ardocrat:
`reqwest` lib:
- use native tls
- disable default features

### DIFF
--- a/impls/Cargo.toml
+++ b/impls/Cargo.toml
@@ -22,7 +22,7 @@ ring = "0.17.14"
 chrono = { version = "0.4.11", features = ["serde"] }
 lazy_static = "1"
 tokio = { version = "1.52.3", features = ["full"] }
-reqwest = { version = "0.13.3", features = ["socks"] }
+reqwest = { version = "0.13.4", default-features = false, features = ["socks", "native-tls"] }
 
 #Socks/Tor/Bridge/Proxy
 ed25519-dalek = "2"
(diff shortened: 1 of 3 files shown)


impls/Cargo.toml
```diff
@@ -22,7 +22,7 @@ ring = "0.17.14"
 chrono = { version = "0.4.11", features = ["serde"] }
 lazy_static = "1"
 tokio = { version = "1.52.3", features = ["full"] }
-reqwest = { version = "0.13.3", features = ["socks"] }
+reqwest = { version = "0.13.4", default-features = false, features = ["socks", "native-tls"] }
```
wiesche89: The default features also removes Reqwest's separate system-proxy feature. The normal HTTPNodeClient::new() path does not set an explicit proxy, so the macOS and Windows system proxy settings would no longer be used. Could we enable system-proxy explicitly? This does not bring back Rustls or rustls-platform-verifier

impls/src/client_utils/client.rs
```diff
@@ -72,7 +72,7 @@ impl Client {
 
 		let mut builder = ClientBuilder::new()
 			.timeout(request_timeout)
-			.use_rustls_tls()
+			.use_native_tls()
```
wiesche89: Reqwest documents use_native_tls() as the old alias for tls_backend_native(). Since this line is already changing, could we use the current method directly?

impls/Cargo.toml
ardocrat: bring back default features except tls

impls/src/client_utils/client.rs
ardocrat: removed

wiesche89 (approved):

Issue #472, opened 2020-07-14
Serialization Error with Large Amounts of Inputs or balance too high


quentinlesceller:
Sender side:
```
grin-wallet -p "" send -d tgrin1m7vmg0z5tu373haxtglape4hv45lw86xscq93ur7jj5rmpk5sw7qf2xf35 -f 1000000

20200714 17:30:02.439 INFO grin_util::logger - log4rs is initialized, file level: Debug, stdout level: Debug, min. level: Debug
20200714 17:30:02.439 INFO grin_wallet - Using wallet configuration file at /Users/quentin/test/grin-wallet.toml
20200714 17:30:02.439 INFO grin_wallet - This is Grin Wallet version 4.0.0, built for x86_64-apple-darwin by rustc 1.44.0.
20200714 17:30:02.439 DEBUG grin_wallet - Built with profile "release", features "".
20200714 17:30:02.888 DEBUG grin_store::lmdb - DB Mapsize for /Users/quentin/test/wallet_data/db/lmdb is 402653184
20200714 17:30:02.890 DEBUG grin_wallet_impls::lifecycle::seed - Using wallet seed file at: /Users/quentin/test/wallet_data/wallet.seed
20200714 17:30:04.478 DEBUG grin_wallet_libwallet::internal::updater - Refreshing wallet outputs
20200714 17:30:34.984 DEBUG grin_wallet_libwallet::internal::selection - Extending maximum number of outputs. 16670 outputs selected.
20200714 17:30:35.070 DEBUG grin_wallet_libwallet::internal::selection - Building change outputs: total change: 43169001078 (1 outputs)
20200714 17:30:35.084 DEBUG grin_core::libtx::build - Building output: 43169001078, Commitment(09b066b489a7271af95bb0a6c8658f7a9a91ac5791e2b875226d6148114d99a3bc)
20200714 17:30:35.408 INFO grin_wallet_controller::command - Tx created: 1000000.000000000 grin to tgrin1m7vmg0z5tu373haxtglape4hv45lw86xscq93ur7jj5rmpk5sw7qf2xf35 (strategy 'smallest')
20200714 17:30:35.408 WARN grin_wallet_api::owner - Attempting to send transaction via TOR
20200714 17:30:35.408 INFO grin_wallet_impls::adapters::http - Starting TOR Process for send at Some(V4(127.0.0.1:59050))
[...]

PR #780, opened 2026-07-27
Optimize outputs scan and selection

ardocrat:
- commit outputs under single batch with chunks saving on chain commitment instead of reconstructing a regular one for speed-up
- parallel iteration over `proof::rewind` checks
- bypass read limit to read data from database in chunks (fixes https://github.com/mimblewimble/grin-wallet/issues/472)
- optimize outputs selection
- update wallet state before send and process invoice
- calculate right max amount to send to not overtake tx weight
- owner api/rpc calls to calculate max amount to send with fee and max inputs size

### DIFF
--- a/api/src/owner.rs
+++ b/api/src/owner.rs
@@ -473,7 +473,7 @@ where
 		)
 	}
 
-	/// Returns max available value to send for provided selection strategy.
+	/// Returns max available value to send for selecting all outputs.
 	///
 	/// # Arguments
 	/// * `keychain_mask` - Wallet secret mask to XOR against the stored wallet seed before using, if
@@ -495,7 +495,7 @@ where
 	/// argument was set to `true`).
 	/// * The second `u64` is amount to send.
 	/// * The third element is fee
-	/// * The forth is maximum input count for selected strategy
+	/// * The fourth is maximum input count
 	///
 	/// # Example
 	/// Set up as in [`new`](struct.Owner.html#method.new) method above.
--- a/api/src/owner.rs
+++ b/api/src/owner.rs
@@ -473,6 +473,69 @@ where
 		)
 	}
 
+	/// Returns max available value to send for provided selection strategy.
+	///
+	/// # Arguments
+	/// * `keychain_mask` - Wallet secret mask to XOR against the stored wallet seed before using, if
+	/// being used.
+	/// * `refresh_from_node` - If true, the wallet will attempt to contact
+	/// a node (via the [`NodeClient`](../grin_wallet_libwallet/types/trait.NodeClient.html)
+	/// provided during wallet instantiation). If `false`, the results will
+	/// contain transaction information that may be out-of-date (from the last time
+	/// the wallet's output set was refreshed against the node).
+	/// Note this setting is ignored if the updater process is running via a call to
+	/// [`start_updater`](struct.Owner.html#method.start_updater)
+	/// * `minimum_confirmations` - The minimum number of confirmations an output
+	/// should have before it's included in the 'amount_currently_spendable' total
+	///
+	/// # Returns
+	/// * `(bool, u64, u64, u32)` - A tuple:
+	/// * The first `bool` element indicates whether the data was successfully
+	/// refreshed from the node (note this may be false even if the `refresh_from_node`
+	/// argument was set to `true`).
+	/// * The second `u64` is amount to send.
+	/// * The third element is fee
+	/// * The forth is maximum input count for selected strategy
+	///
+	/// # Example
+	/// Set up as in [`new`](struct.Owner.html#method.new) method above.
+	/// ```
+	/// # grin_wallet_api::doctest_helper_setup_doc_env!(wallet, wallet_config);
+	///
+	/// let api_owner = Owner::new(wallet.clone(), None, std::path::PathBuf::from("grin-wallet.toml"));
+	/// let selection_strategy_is_use_all = false;
+	///
+	/// let result = api_owner.estimate_max_sendable(selection_strategy_is_use_all);
+	///
+	/// if let Ok((was_updated, output_mappings)) = result {
+	///     //...
+	/// }
+	/// ```
+
+	pub fn estimate_max_sendable(
+		&self,
+		keychain_mask: Option<&SecretKey>,
+		refresh_from_node: bool,
+		minimum_confirmations: u64,
+	) -> Result<(bool, u64, u64, u32), Error> {
+		let tx = {
+			let t = self.status_tx.lock();
+			t.clone()
+		};
+		let refresh_from_node = match self.updater_running.load(Ordering::Relaxed) {
+			true => false,
+			false => refresh_from_node,
+		};
+
+		owner::estimate_max_sendable(
+			self.wallet_inst.clone(),
+			keychain_mask,
+			&tx,
+			refresh_from_node,
+			minimum_confirmations,
+		)
+	}
+
 	/// Returns a list of [Transaction Log Entries](../grin_wallet_libwallet/types/struct.TxLogEntry.html)
 	/// from the active account in the wallet.
 	///
--- a/api/src/owner.rs
+++ b/api/src/owner.rs
@@ -507,7 +507,7 @@ where
 	/// let minimum_confirmations = 10;
 	/// let result = api_owner.estimate_max_sendable(None, update_from_node, minimum_confirmations);
 	///
-	/// if let Ok((was_updated, output_mappings)) = result {
+	/// if let Ok((was_updated, amount, fee, max_inputs)) = result {
 	///     //...
 	/// }
 	/// ```
--- a/api/src/owner.rs
+++ b/api/src/owner.rs
@@ -503,9 +503,9 @@ where
 	/// # grin_wallet_api::doctest_helper_setup_doc_env!(wallet, wallet_config);
 	///
 	/// let api_owner = Owner::new(wallet.clone(), None, std::path::PathBuf::from("grin-wallet.toml"));
-	/// let selection_strategy_is_use_all = false;
-	///
-	/// let result = api_owner.estimate_max_sendable(selection_strategy_is_use_all);
+	/// let update_from_node = true;
+	/// let minimum_confirmations = 10;
+	/// let result = api_owner.estimate_max_sendable(None, update_from_node, minimum_confirmations);
 	///
 	/// if let Ok((was_updated, output_mappings)) = result {
 	///     //...
--- a/api/src/owner.rs
+++ b/api/src/owner.rs
@@ -604,21 +604,34 @@ where
 		refresh_from_node: bool,
 		minimum_confirmations: u64,
 	) -> Result<(bool, WalletInfo), Error> {
+		let (updated, _, info) = self.retrieve_summary_info_with_refresh_status(
+			keychain_mask,
+			refresh_from_node,
+			minimum_confirmations,
+		)?;
+		Ok((updated, info))
+	}
+
+	/// Retrieve summary info and whether refresh was skipped.
+	pub fn retrieve_summary_info_with_refresh_status(
+		&self,
+		keychain_mask: Option<&SecretKey>,
+		refresh_from_node: bool,
+		minimum_confirmations: u64,
+	) -> Result<(bool, bool, WalletInfo), Error> {
 		let tx = {
 			let t = self.status_tx.lock();
 			t.clone()
 		};
-		let refresh_from_node = match self.updater_running.load(Ordering::Relaxed) {
-			true => false,
-			false => refresh_from_node,
-		};
-		owner::retrieve_summary_info(
+		let refresh_skipped = refresh_from_node && self.updater_running.load(Ordering::Relaxed);
+		let (updated, info) = owner::retrieve_summary_info(
 			self.wallet_inst.clone(),
 			keychain_mask,
 			&tx,
-			refresh_from_node,
+			refresh_from_node && !refresh_skipped,
 			minimum_confirmations,
-		)
+		)?;
+		Ok((updated, refresh_skipped, info))
 	}
 
 	/// Initiates a new transaction as the sender, creating a new
--- a/api/src/owner_rpc.rs
+++ b/api/src/owner_rpc.rs
@@ -229,7 +229,7 @@ pub trait OwnerRpc {
 	) -> Result<(bool, Vec<OutputCommitMapping>), Error>;
 
 	/**
-	Networked version of [Owner::retrieve_txs](struct.Owner.html#method.estimate_max_sendable).
+	Networked version of [Owner::estimate_max_sendable](struct.Owner.html#method.estimate_max_sendable).
 
 	# Json rpc example
 
--- a/api/src/owner_rpc.rs
+++ b/api/src/owner_rpc.rs
@@ -461,62 +461,62 @@ pub trait OwnerRpc {
 	) -> Result<(bool, WalletInfo), Error>;
 
 	/**
-		;Networked version of [Owner::init_send_tx](struct.Owner.html#method.init_send_tx).
-	z
-		```
-			# grin_wallet_api::doctest_helper_json_rpc_owner_assert_response!(
-			# r#"
-			{
-				"jsonrpc": "2.0",
-				"method": "init_send_tx",
-				"params": {
-					"token": "d202964900000000d302964900000000d402964900000000d502964900000000",
-					"args": {
-						"src_acct_name": null,
-						"amount": "6000000000",
-						"minimum_confirmations": 2,
-						"max_outputs": 500,
-						"num_change_outputs": 1,
-						"selection_strategy_is_use_all": true,
-						"refresh_outputs_from_node": true,
-						"target_slate_version": null,
-						"payment_proof_recipient_address": "tgrin1xtxavwfgs48ckf3gk8wwgcndmn0nt4tvkl8a7ltyejjcy2mc6nfs9gm2lp",
-						"ttl_blocks": null,
-						"send_args": null
-					}
-				},
-				"id": 1
-			}
-			# "#
-			# ,
-			# r#"
-			{
-			"id": 1,
-				"jsonrpc": "2.0",
-				"result": {
-					"Ok": {
-						"amt": "6000000000",
-						"fee": "23000000",
-						"id": "0436430c-2b02-624c-2032-570501212b00",
-						"proof": {
-							"raddr": "32cdd63928854f8b2628b1dce4626ddcdf35d56cb7cfdf7d64cca5822b78d4d3",
-							"saddr": "32cdd63928854f8b2628b1dce4626ddcdf35d56cb7cfdf7d64cca5822b78d4d3"
-						},
-						"sigs": [
-							{
-								"nonce": "031b84c5567b126440995d3ed5aaba0565d71e1834604819ff9c17f5e9d5dd078f",
-								"xs": "02e89cce4499ac1e9bb498dab9e3fab93cc40cd3d26c04a0292e00f4bf272499ec"
-							}
-						],
-						"sta": "S1",
-						"ver": "4:2"
-					}
+	Networked version of [Owner::init_send_tx](struct.Owner.html#method.init_send_tx).
+
+	```
+		# grin_wallet_api::doctest_helper_json_rpc_owner_assert_response!(
+		# r#"
+		{
+			"jsonrpc": "2.0",
+			"method": "init_send_tx",
+			"params": {
+				"token": "d202964900000000d302964900000000d402964900000000d502964900000000",
+				"args": {
+					"src_acct_name": null,
+					"amount": "6000000000",
+					"minimum_confirmations": 2,
+					"max_outputs": 500,
+					"num_change_outputs": 1,
+					"selection_strategy_is_use_all": true,
+					"refresh_outputs_from_node": true,
+					"target_slate_version": null,
+					"payment_proof_recipient_address": "tgrin1xtxavwfgs48ckf3gk8wwgcndmn0nt4tvkl8a7ltyejjcy2mc6nfs9gm2lp",
+					"ttl_blocks": null,
+					"send_args": null
+				}
+			},
+			"id": 1
+		}
+		# "#
+		# ,
+		# r#"
+		{
+		"id": 1,
+			"jsonrpc": "2.0",
+			"result": {
+				"Ok": {
+					"amt": "6000000000",
+					"fee": "23000000",
+					"id": "0436430c-2b02-624c-2032-570501212b00",
+					"proof": {
+						"raddr": "32cdd63928854f8b2628b1dce4626ddcdf35d56cb7cfdf7d64cca5822b78d4d3",
+						"saddr": "32cdd63928854f8b2628b1dce4626ddcdf35d56cb7cfdf7d64cca5822b78d4d3"
+					},
+					"sigs": [
+						{
+							"nonce": "031b84c5567b126440995d3ed5aaba0565d71e1834604819ff9c17f5e9d5dd078f",
+							"xs": "02e89cce4499ac1e9bb498dab9e3fab93cc40cd3d26c04a0292e00f4bf272499ec"
+						}
+					],
+					"sta": "S1",
+					"ver": "4:2"
 				}
 			}
-			# "#
-			# , 4, false, false, false, false);
-		```
-		*/
+		}
+		# "#
+		# , 4, false, false, false, false);
+	```
+	*/
 
 	fn init_send_tx(&self, token: Token, args: InitTxArgs) -> Result<VersionedSlate, Error>;
 
--- a/api/src/owner_rpc.rs
+++ b/api/src/owner_rpc.rs
@@ -461,62 +461,62 @@ pub trait OwnerRpc {
 	) -> Result<(bool, WalletInfo), Error>;
 
 	/**
-	;Networked version of [Owner::init_send_tx](struct.Owner.html#method.init_send_tx).
-
-	```
-		# grin_wallet_api::doctest_helper_json_rpc_owner_assert_response!(
-		# r#"
-		{
-			"jsonrpc": "2.0",
-			"method": "init_send_tx",
-			"params": {
-				"token": "d202964900000000d302964900000000d402964900000000d502964900000000",
-				"args": {
-					"src_acct_name": null,
-					"amount": "6000000000",
-					"minimum_confirmations": 2,
-					"max_outputs": 500,
-					"num_change_outputs": 1,
-					"selection_strategy_is_use_all": true,
-					"refresh_outputs_from_node": true,
-					"target_slate_version": null,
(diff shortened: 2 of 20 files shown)


libwallet/src/internal/scan.rs
```diff
@@ -75,49 +64,69 @@ fn identify_utxo_outputs<K>(
...
-		// will fail if it's not ours
-		let info = {
-			// Before HF+2wk, try legacy rewind first
-			let info_legacy =
-				if valid_header_version(height.saturating_sub(2 * WEEK_HEIGHT), legacy_version) {
-					proof::rewind(keychain.secp(), &legacy_builder, *commit, None, *proof)?
-				} else {
-					None
+	let identified = outputs
+		.par_iter()
```
wiesche89, line 74: This moves chain dependent calls onto Rayon workers, which do not inherit set_local_chain_type(). Without a global chain type, the scan can panic. Please capture the chain type before the loop and set it locally inside each worker.

libwallet/src/internal/scan.rs
```diff
@@ -285,12 +285,15 @@ where
 	Ok((result_vec, last_retrieved_return_index, perc_complete))
 }
 
-fn restore_missing_output<'a, L, C, K>(
+const MISSING_OUTPUTS_BATCH_SIZE: usize = 10000;
```
wiesche89: A 10k write batch can hit MDB_MAP_FULL because LMDB only resizes before opening the transaction. I reproduced this with the test map. 1000 works safely. Please use the smaller batch size.

libwallet/src/api_impl/owner_updater.rs
```diff
@@ -72,8 +73,11 @@ pub fn start_updater_log_thread(
 					StatusMessage::UpdatingTransactions(s) => debug!("{}", s),
 					StatusMessage::FullScanWarn(s) => warn!("{}", s),
 					StatusMessage::Scanning(s, m) => {
-						debug!("{}", s);
-						warn!("Scanning - {}% complete", m);
+						if last_scan_percent < m {
```
wiesche89, line 80: This also hides different debug messages with the same percentage, and the value is never reset for later scans. Please always log s and only deduplicate the percentage warning.

libwallet/src/types.rs
```diff
@@ -888,6 +888,22 @@ impl ViewWalletOutputResult {
 	}
 }
 
+fn read_bytes_len_prefix<R: ser::Reader>(reader: &mut R) -> Result<Vec<u8>, ser::Error> {
+	let mut len = reader.read_u64()? as usize;
+	let limit = reader.read_limit().unwrap();
```
wiesche89: This panics when the reader has no limit, and zero would cause an endless loop. It also removes the total size limit for several wallet types. Please handle None and zero, and keep a wallet-specific total limit for Context.

libwallet/src/types.rs
```diff
@@ -490,7 +490,7 @@ impl ser::Writeable for Context {
 
 impl ser::Readable for Context {
 	fn read<R: ser::Reader>(reader: &mut R) -> Result<Context, ser::Error> {
-		let data = reader.read_bytes_len_prefix()?;
+		let data = read_bytes_len_prefix(reader)?;
```
wiesche89, line 501: Could we add a regression test with a serialized Context larger than 100 KB? Around 2.000 inputs are enough to reproduce the original issue.

libwallet/src/internal/scan.rs
```diff
@@ -299,72 +302,56 @@ where
 {
 	wallet_lock!(wallet_inst, w);
```
wiesche89: The wallet lock is held across all chunks, so a large restore blocks every other wallet operation. Could we acquire it per batch instead?

libwallet/src/internal/scan.rs
```diff
@@ -299,72 +302,56 @@ where
...
+			t.num_outputs = 1;
+			t.update_confirmation_ts();
+			batch.save_tx_log_entry(t, &parent_key_id)?;
+
+			batch.save(OutputData {
+				root_key_id: parent_key_id.clone(),
+				key_id: output.key_id.clone(),
+				n_child: output.n_child,
+				mmr_index: Some(output.mmr_index),
+				commit: Some(output.commit.0.to_vec().to_hex()),
```
wiesche89, line 343: This now stores the real on chain commitment instead of reconstructing a regular one. That looks correct, but it is a behavioral change and should be mentioned in the PR description or covered by a test.

libwallet/src/internal/scan.rs
```diff
@@ -299,72 +302,56 @@ where
...
-				parent_key_id.clone(),
-				RestoredTxStats {
-					log_id: batch.next_tx_log_id(&parent_key_id)?,
-					amount_credited: 0,
-					num_outputs: 0,
-				},
-			);
-		}
-	}
+	for o_chunks in outputs.chunks(MISSING_OUTPUTS_BATCH_SIZE).into_iter() {
```
wiesche89: chunks() is already iterable, so .into_iter() can be removed. It currently adds a Clippy warning.

api/src/owner_rpc.rs
```diff
@@ -1123,54 +1123,54 @@ pub trait OwnerRpc {
...
-			 "method": "encrypted_request_v3",
-			 "id": "1",
-			 "params": {
-					"nonce": "ef32...",
-					"body_enc": "e0bcd..."
-			 }
-		}
-		# "#;
-		```
+	Initializes the secure JSON-RPC API. This function must be called and a shared key
```
wiesche89, line 1173: This block only changes documentation indentation and is unrelated to the scan optimization. Could we revert it to keep the PR focused?
Or was this just reformatted by cargo fmt?

libwallet/src/internal/scan.rs
ardocrat: yes, forgot to remove it

api/src/owner_rpc.rs
ardocrat: this is some formatting change, documentation was not changed

libwallet/src/internal/scan.rs
ardocrat: this did speed improvement in comparison, need to comment

libwallet/src/internal/scan.rs
ardocrat: fixed

libwallet/src/types.rs
ardocrat: fixed

libwallet/src/api_impl/owner_updater.rs
ardocrat: fixed

libwallet/src/internal/scan.rs
ardocrat: fixed

libwallet/src/internal/scan.rs
ardocrat: added

libwallet/src/types.rs
ardocrat: added

libwallet/src/types.rs
```diff
@@ -888,6 +888,26 @@ impl ViewWalletOutputResult {
 	}
 }
 
+fn read_bytes_len_prefix<R: ser::Reader>(reader: &mut R) -> Result<Vec<u8>, ser::Error> {
+	let mut len = reader.read_u64()? as usize;
+	match reader.read_limit() {
+		None => reader.read_bytes_len_prefix(),
```
wiesche89: Thanks for the large Context test. The None path still reads the prefix twice with a StreamingReader this reproducibly aborts while allocating. Some(0) also never terminates. Could we use read_fixed_bytes(len) for None and reject a zero limit?

libwallet/src/api_impl/owner_updater.rs
```diff
@@ -73,7 +74,10 @@ pub fn start_updater_log_thread(
 					StatusMessage::FullScanWarn(s) => warn!("{}", s),
 					StatusMessage::Scanning(s, m) => {
 						debug!("{}", s);
-						warn!("Scanning - {}% complete", m);
+						if last_scan_percent < m {
+							warn!("Scanning - {}% complete", m);
```
wiesche89, line 81: The debug output looks good now, thanks. last_scan_percent remains at 99 after ScanningComplete, so later scans no longer log progress. Could we reset it when a scan completes?

controller/src/command.rs
```diff
@@ -351,9 +351,9 @@ where
 
 	let mut slate = Slate::blank(2, false);
 	let mut amount = args.amount;
+	let (_, wallet_info) =
+		owner_api.retrieve_summary_info(keychain_mask, true, args.minimum_confirmations)?;
```
wiesche89: The refresh can return (false, cached_info) when the node is unavailable, and it is skipped while the updater is running. Since the result is ignored here, send/invoice may still use stale state. Could we detect the best effort case and warn the user?

ardocrat: `command::send` already returns an error if node is not available as we are requesting `get_tip`:

image

will add additional logging for `owner_api.retrieve_summary_info`

libwallet/src/api_impl/owner_updater.rs
ardocrat: added

libwallet/src/types.rs
ardocrat: fixed, 0 limit can be understood as no limit

controller/src/command.rs
```diff
@@ -351,11 +351,14 @@ where
 
 	let mut slate = Slate::blank(2, false);
 	let mut amount = args.amount;
+	let (info_updated, wallet_info) =
+		owner_api.retrieve_summary_info(keychain_mask, true, args.minimum_confirmations)?;
```
wiesche89: This refresh starts one request per 500 outputs without a limit. Could we bound the concurrency?

controller/src/command.rs
```diff
@@ -351,11 +351,14 @@ where
...
 	if args.use_max_amount {
-		let (_, wallet_info) =
-			owner_api.retrieve_summary_info(keychain_mask, true, args.minimum_confirmations)?;
 		amount = wallet_info.amount_currently_spendable;
 	}
+	if !info_updated {
```
wiesche89: While the updater is running, the refresh is intentionally skipped and returns false. This makes every interactive send/pay warn even when the node is healthy. Could we distinguish “skipped” from “failed”?

libwallet/src/types.rs
```diff
@@ -888,6 +888,29 @@ impl ViewWalletOutputResult {
 	}
 }
 
+fn read_bytes_len_prefix<R: ser::Reader>(reader: &mut R) -> Result<Vec<u8>, ser::Error> {
```
wiesche89, line 899: None and zero are fixed, thanks. However, chunking removes the total bound for six database types. Could we keep it Context-specific and enforce the same 4 MB limit on reads and writes?

libwallet/src/types.rs
```diff
@@ -966,4 +991,25 @@ mod tests {
...
+		}
+		let ser_value = ser::ser_vec(&context, ProtocolVersion(3)).unwrap();
+		let mut value = ser_value.as_slice();
+		assert!(value.len() > 100_000);
+		let context = ser::deserialize::<Context, &[u8]>(
+			&mut value,
+			ProtocolVersion(3),
+			DeserializationMode::Full,
+		);
+		assert!(context.is_ok());
```
wiesche89: This only uses BinReader, so the fixed StreamingReader None path remains untested. Could we read the same Context through it and then verify all 3000 outputs?

libwallet/src/types.rs
ardocrat: `TxLogEntry` can be also big if we will keep `stored_tx` here. 4Mb is something specific for tx weight, not sure if should put it here.

controller/src/command.rs
ardocrat: Max amount of outputs will be added by checking `Transaction::weight_by_iok` on send tx creation

controller/src/command.rs
ardocrat: changed

controller/src/command.rs
ardocrat: 500 can lead to problems for large requests, not is getting "too much open files" error. It makes sense to specify max amount of requests or remove chunks at all.

libwallet/src/types.rs
ardocrat: added test

libwallet/src/types.rs
ardocrat: keeping limit check only for context and tx log entry

controller/src/command.rs
ardocrat: sending 16 req to node

impls/src/node_clients/http.rs
```diff
@@ -259,17 +259,27 @@ impl NodeClient for HTTPNodeClient {
...
-					url.as_str(),
-					req,
-					api_secret.clone(),
-				));
+			let mut outputs = Vec::new();
+
+			let max_num_requests = 16;
+			for req_chunks in reqs.chunks(max_num_requests) {
+				for req in req_chunks {
+					let mut tasks = vec![];
```
wiesche89: This currently runs fully serially, each FuturesUnordered contains one request and is drained immediately. Could we build it from the whole 16 request chunk and then await the batch?

libwallet/src/internal/selection.rs
```diff
@@ -113,6 +115,19 @@ where
 		);
 	}
 
+	let input_len = inputs.len();
+	let output_len = change_amounts_derivations.len();
+	let tx_weight = Transaction::weight_by_iok(input_len as u64, output_len as u64, 1u64);
```
wiesche89: This misses the final recipient output. At the weight boundary the check can accept a transaction that the node later rejects. Could we use output_len + 1?

controller/src/command.rs
wiesche89: init_send_tx() calls refresh_outputs() again afterwards, so large wallets perform all output requests twice. Could we reuse the first refresh?

libwallet/src/types.rs
```diff
@@ -966,4 +991,30 @@ mod tests {
...
+		}
+		let ser_value = ser::ser_vec(&context, protocol_ver).unwrap();
+		let mut value = ser_value.as_slice();
+		assert!(value.len() > 100_000);
+		let context =
+			ser::deserialize::<Context, &[u8]>(&mut value, protocol_ver, DeserializationMode::Full);
+		assert!(context.is_ok());
+
+		// Big read for StreamingReader returns an error, cause limit is not set.
+		let mut streaming_reader = StreamingReader::new(&mut value, protocol_ver);
```
wiesche89, line 1040: value was already consumed above, so this only tests EOF. Could we use a fresh slice, call Context::read(), and then verify the 3000 outputs?

controller/src/command.rs
wiesche89: false still means either “the updater is running” or “the node refresh failed.” Could we distinguish the reason instead of reporting both as “skipped”?

libwallet/src/api_impl/owner_updater.rs
wiesche89: A failed scan never sends ScanningComplete, so its retry suppresses progress up to the old value. Could we also reset when a new 0% scan starts?

libwallet/src/types.rs
wiesche89: This still limits individual reads, not the total length. stored_tx is also just a filename. Could we keep this Context-specific and enforce the same 4 MiB limit on reads and writes?

controller/src/command.rs
wiesche89: init_send_tx() calls refresh_outputs() again afterwards, so large wallets perform all output requests twice. Could we reuse the first refresh?

libwallet/src/types.rs
ardocrat: I meant for the future improvement, added read_bytes_len_prefix

controller/src/command.rs
ardocrat: added actual reason

libwallet/src/internal/selection.rs
ardocrat: added

libwallet/src/types.rs
ardocrat: added

controller/src/command.rs
ardocrat: then we need to provide flag to Owner::init_send_tx() to refresh outputs or not

libwallet/src/api_impl/owner_updater.rs
ardocrat: added

impls/src/node_clients/http.rs
ardocrat: added tasks for chunk

controller/src/command.rs
ardocrat: added flag to initsendtx

wiesche89 (approved):

libwallet/src/api_impl/owner.rs
```diff
@@ -293,6 +294,61 @@ where
...
+		max_outputs,
+		change_outputs,
+		true,
+		&parent_key_id,
+	) {
+		Ok((coins, _total, amount, fee)) => (amount, fee, coins.len() as u32),
+		Err(e) => match e {
+			Error::BigAmountError(amount, input_count) => {
+				let fee = tx_fee(input_count as usize, change_outputs + 1, 1);
+				(amount, fee, input_count)
```
wiesche89, line 347: The two branches return different semantics, on success, amount is already net, but here it is still the gross sum because amount_includes_fee is true, while the fee is returned separately. A client therefore cannot send this value as documented. Could we subtract the fee here as well and add an overweight-path test?

controller/src/command.rs
```diff
@@ -351,27 +366,53 @@ where
...
+			true,
+			args.minimum_confirmations,
+		)?;
 	if args.use_max_amount {
-		let (_, wallet_info) =
-			owner_api.retrieve_summary_info(keychain_mask, true, args.minimum_confirmations)?;
 		amount = wallet_info.amount_currently_spendable;
 	}
+	if !info_updated && !update_skipped {
+		warn!("Wallet info update was skipped: node connection error");
```
wiesche89: Sorry, I made a mistake here yesterday. This condition means the refresh failed rather than being skipped. Could we change the message to Wallet info update failed: node connection error? The same applies to the second block below.

api/src/owner.rs
```diff
@@ -473,6 +473,69 @@ where
 		)
 	}
 
+	/// Returns max available value to send for provided selection strategy.
```
wiesche89: This method does not accept a selection strategy and always uses all internally. Could we document that explicitly? Also, forth below should be fourth.

api/src/owner_rpc.rs
```diff
@@ -228,6 +228,51 @@ pub trait OwnerRpc {
 		tx_id: Option<u32>,
 	) -> Result<(bool, Vec<OutputCommitMapping>), Error>;
 
+	/**
+	Networked version of [Owner::retrieve_txs](struct.Owner.html#method.estimate_max_sendable).
```
wiesche89: The link text should be Owner::estimate_max_sendable, not Owner::retrieve_txs.

libwallet/src/api_impl/owner.rs
ardocrat: > Could we subtract the fee here

added

ardocrat: added test for bitamounterror for max send

libwallet/src/api_impl/owner.rs
```diff
@@ -293,6 +293,58 @@ where
...
+		current_height,
+		minimum_confirmations,
+		max_outputs,
+		change_outputs,
+		true,
+		&parent_key_id,
+	) {
+		Ok((coins, _total, amount, fee)) => (amount, fee, coins.len() as u32),
+		Err(e) => match e {
+			Error::BigAmountError(amount, fee, input_count) => (amount - fee, fee, input_count),
```
wiesche89: There is still an edge case here. BigAmountError does not guarantee that amount is greater than fee. With enough tiny outputs, this subtraction can underflow. Could we use checked_sub as in select_coins_and_fee and add a low value output test?

controller/src/command.rs
```diff
@@ -351,27 +366,58 @@ where
...
-		let strategies = vec!["smallest", "all"]
-			.into_iter()
+		let strategies = estimate_strategies(args.use_max_amount)
+			.iter()
+			.copied()
 			.map(|strategy| {
-				let init_args = InitTxArgs {
+				let (_, a, fee, inputs) = owner_api
+					.estimate_max_sendable(keychain_mask, !info_updated, args.minimum_confirmations)
+					.unwrap();
```
wiesche89: This max estimate is only logged and does not affect the result. unwrap() can panic here, for example on an empty wallet, and the selection runs once per strategy. Could we remove this block?

libwallet/src/error.rs
```diff
@@ -37,6 +37,13 @@ pub enum Error {
 		needed_disp: String,
 	},
 
+	/// Big amount error.
+	#[error(
+		"Amount too large for a single transaction; can send at most {} using {2} inputs. Send that amount to yourself using the 'all' selection strategy and a maximum of {1} inputs to consolidate outputs",
```
wiesche89: The second input placeholder still uses {1}, which is now the fee. This can print the fee as the maximum input count. Could we use {2} here as well and check both occurrences in the test?

libwallet/src/internal/selection.rs
```diff
@@ -476,6 +478,30 @@ where
...
+			&coins,
+			output_len as u64,
+			max_tx_weight,
+			amount_includes_fee,
+		);
+		error!(
+			"Transaction weight {}, exceeds global max_tx_weight {}, can send maximum {}, send such amount to yourself for outputs consolidation",
+			tx_weight, max_tx_weight, amount_to_hr_string(max_amount, true)
+		);
+		let fee = tx_fee(max_inputs as usize, output_len, 1);
```
wiesche89: The suggested consolidation amount is based on one output, but this fee still uses the failed transaction’s output_len. Could we calculate the error fee with one output as well and cover the change output case?

libwallet/src/lib.rs
```diff
@@ -72,6 +72,7 @@ pub use api_impl::types::{
 };
 pub use backend::{WalletBackend, WalletBatch};
 pub use internal::scan::scan;
+pub use internal::selection::select_coins_and_fee;
```
wiesche89: This internal function is only exported for the new controller test. Could the test trigger the error through api.init_send_tx, so the helper can stay private and the real public path is covered?

controller/src/command.rs
```diff
@@ -351,27 +366,58 @@ where
...
+			true,
+			args.minimum_confirmations,
+		)?;
 	if args.use_max_amount {
-		let (_, wallet_info) =
-			owner_api.retrieve_summary_info(keychain_mask, true, args.minimum_confirmations)?;
 		amount = wallet_info.amount_currently_spendable;
 	}
+	if !info_updated && !update_skipped {
+		warn!("Wallet info update was failed: node connection error");
```
wiesche89: Small correction: this should be "Wallet info update failed: node connection error". The same applies to the block below.

controller/src/command.rs
ardocrat: this added for test, removed

libwallet/src/error.rs
ardocrat: fixed

libwallet/src/internal/selection.rs
ardocrat: added 1 output here

libwallet/src/api_impl/owner.rs
ardocrat: it means we should calculate right max amount to prevent such issue at `selection::max_spendable_amount`

libwallet/src/lib.rs
ardocrat: fixed test

libwallet/src/api_impl/owner.rs
ardocrat: also added check at owner

libwallet/src/internal/selection.rs
```diff
@@ -487,6 +513,57 @@ where
...
+	let mut total_value = 0;
+	let (amount, inputs) = values
+		.into_iter()
+		.enumerate()
+		.take_while(|(index, value)| {
+			let inputs_len = *index as u64 + 1;
+			let fee = tx_fee(inputs_len as usize, 1, 1);
+			total_value += value;
+			Transaction::weight_by_iok(inputs_len, output_len, 1) <= max_tx_weight
+				&& total_value > fee
```
wiesche89: Selecting the smallest outputs first is intentional. However, this condition stops take_while when the first output alone does not cover the fee, even though later weight eligible inputs would make the total sufficient. Could we determine the weight limited prefix first and then check its total against the fee?

libwallet/src/error.rs
```diff
@@ -367,3 +375,17 @@ impl From<bech32::Error> for Error {
...
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn big_amount_message() {
+		let message = Error::BigAmountError(3_000_000_000, 12_500_000, 2).to_string();
+		assert!(message.contains("3.0"));
+		assert!(message.contains("2 inputs"));
```
wiesche89: The message is correct now, but this assertion only checks that "2 inputs" appears at least once. If either placeholder regresses to {1}, the test still passes. Could we verify both occurrences or assert the full message?

ardocrat: check 2 times to appear

libwallet/src/internal/selection.rs
ardocrat: added sort to decrease value at next iteration

libwallet/src/internal/selection.rs
```diff
@@ -487,6 +513,66 @@ where
...
+
+	let mut total_value = 0;
+	let (amount, inputs) = values
+		.into_iter()
+		.enumerate()
+		.take_while(|(index, value)| {
+			let inputs_len = *index as u64 + 1;
+			let fee = tx_fee(inputs_len as usize, 1, 1);
+			total_value += value;
+			total_value > fee
```
wiesche89: The sort helps when one output already covers the base fee, but take_while still stops when multiple outputs are needed to cover it together. For example, [20, 20] currently returns (0, 0) even though the total 40 exceeds the two input fee 26. Could we check the fee after summing the selected inputs?

libwallet/src/internal/selection.rs
```diff
@@ -487,6 +513,66 @@ where
...
+		.map(|output| output.value)
+		.enumerate()
+		.take_while(|(index, _)| {
+			Transaction::weight_by_iok(*index as u64 + 1, output_len, 1) <= max_tx_weight
+		})
+		.map(|(_, value)| value)
+		.collect();
+
+	// sort outputs by decreasing value to calculate the best fee
+	values.sort_by_key(|out| -(*out as i64));
```
wiesche89: Since these values are u64, casting to i64 can misorder large values and negating i64::MIN can overflow.

PR #782, opened 2026-07-28
Make updater apis public

ardocrat:
Make `updater` APIs public at `grin_wallet_libwallet` for better integration and outputs control, for example we need to calculate balances for all accounts (accounts API can be extented for this).

- fix: provide parent key to get last confirmed height for right account

### DIFF
--- a/libwallet/src/internal/updater.rs
+++ b/libwallet/src/internal/updater.rs
@@ -402,6 +402,7 @@ where
 
 /// Refreshes the outputs in a wallet with the latest information
 /// from a node
+/// Also removes stale unconfirmed coinbase outputs across all accounts
 pub fn refresh_outputs<C, K>(
 	wallet: &mut WalletBackend<C, K>,
 	keychain_mask: Option<&SecretKey>,
--- a/libwallet/src/internal/updater.rs
+++ b/libwallet/src/internal/updater.rs
@@ -420,7 +420,7 @@ where
 /// Build a local map of wallet outputs keyed by commit
 /// and a list of outputs we want to query the node for.
 /// If `update_all` equals `false` we will select outputs
-/// that are actually involved in an outstanding transaction.
+/// that are actually involved in existing transactions for account.
 /// Returns mapping of output commit to tuple of derived key for output,
 /// PMMR index, tx entry log identifier and check if output is unspent
 pub fn map_wallet_outputs<C, K>(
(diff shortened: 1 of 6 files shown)


libwallet/src/lib.rs
```diff
@@ -72,6 +72,7 @@ pub use api_impl::types::{
 };
 pub use backend::{WalletBackend, WalletBatch};
 pub use internal::scan::scan;
+pub use internal::updater;
```
wiesche89: This exposes all internal updater functions. Do we really want all of them to be public, or should we only export those intended for integrations?

ardocrat: actually all functions with `parent_key_id` arg, common issue is wallet only uses current account to update info from owner and foreign APIs

libwallet/src/lib.rs
```diff
@@ -72,6 +72,10 @@ pub use api_impl::types::{
 };
 pub use backend::{WalletBackend, WalletBatch};
 pub use internal::scan::scan;
+pub use internal::updater::{
+	apply_advanced_tx_list_filtering, cancel_tx_and_outputs, map_wallet_outputs, refresh_outputs,
```
wiesche89: Thanks for narrowing the api. cancel_tx_and_outputs skips the checks in tx::cancel_tx, so could we export the checked function instead? apply_advanced_tx_list_filtering is already used through retrieve_txs and does not need a separate export. If map_wallet_outputs stays public, could we briefly document its tuple fields and update_all?

ardocrat: removed public apply_advanced_tx_list_filtering, cancel_tx_and_outputs, added comment for map_wallet_outputs

libwallet/src/lib.rs
```diff
@@ -72,6 +72,9 @@ pub use api_impl::types::{
 };
 pub use backend::{WalletBackend, WalletBatch};
 pub use internal::scan::scan;
+pub use internal::updater::{
+	map_wallet_outputs, refresh_outputs, retrieve_info, retrieve_outputs, retrieve_txs,
```
wiesche89, line 76: These functions accept a parent_key_id, but retrieve_info and the height check in refresh_outputs still use the active account’s confirmed height. This can return incorrect wallet info or skip an update when another account is requested. Could we read the height for the supplied account and add a test for a non active account?

libwallet/src/internal/updater.rs
```diff
@@ -417,8 +417,12 @@ where
 	Ok(())
 }
 
-/// build a local map of wallet outputs keyed by commit
-/// and a list of outputs we want to query the node for
+/// Build a local map of wallet outputs keyed by commit
```
wiesche89: Thanks for adding the docs. Two small details, with update_all == false, outputs without a tx_log_entry are kept as well, and the function only returns the map the commits to query are derived from its keys. Could you adjust the wording?

ardocrat: extended comment about update_all == false

libwallet/src/lib.rs
ardocrat: fixed retrieve_info, refresh_outputs uses height from node

wiesche89 (approved):

PR #783, opened 2026-07-28
Mwixnet wallet support

wiesche89:
**Summary**
Adds Owner API support for creating signed mwixnet swap requests from wallet outputs.
Related PR: https://github.com/mimblewimble/mwixnet/pull/34

- Adds an experimental mwixnet command that submits requests over Arti. Outputs can be selected by commitment, minimum amount, or max. The selected output is printed and tracked in the transaction log.

- Requests that were definitely not sent are cancelled and unlocked. Rejected or uncertain requests remain locked.

- Also fixes Arti HTTP POST request formatting and error handling.

**Tests**

Adds tests for request creation, locking, output selection, JSON-RPC responses, transaction IDs, and Arti HTTP formatting.

**Testing**
Select an unspent output:
./target/release/grin-wallet --testnet outputs

Select an output by commitment, minimum amount, or max:
./target/release/grin-wallet --testnet mwixnet \
  <OUTPUT_COMMITMENT|MINIMUM_AMOUNT|max> \
  --server mza3u6vkqodqc6kjjapfbcjrg7hgpkuw6nyq5xvfjqp5jm2lfeaocwqd.onion \
  --fee_per_hop 12500000 \
  --key ba2ee29d1e44227e3d78d1abbc7d86a9321ae24fb3656b545b001907a008bc4e \
  --key a6920c1ef8bd8bb14fe4940521eca294f797856baaa307eb51894dda157b1477
  
- OUTPUT_COMMITMENT selects an explicit wallet output.
- MINIMUM_AMOUNT selects the smallest eligible output greater than or equal to that amount.
- max selects the largest eligible output.

The selected commitment and its actual value are printed before submission. 
This automatic selection is CLI only

**Help**
./target/release/grin-wallet --testnet mwixnet --help


controller/src/command.rs
```diff
@@ -327,6 +332,120 @@ pub struct SendArgs {
...
+			Ok(())
+		}
+		MwixnetResponse::Rejected(message) => Err(Error::GenericError(format!(
+			"MWixnet rejected the request: {}. Output remains locked as transaction {}",
+			message, tx_id
+		))),
+	}
+}
+
+pub fn mwixnet<L, C, K>(
```
ardocrat, line 375: question about usability, so user should create send request, extract commitment somehow and provide id to wallet, can we use this command on send to automate commitment getting

controller/src/command.rs
```diff
@@ -327,6 +332,120 @@ pub struct SendArgs {
...
+	pub minimum_confirmations: u64,
+	pub params: MixnetReqCreationParams,
+}
+
+enum MwixnetResponse {
+	Accepted,
+	Rejected(String),
+}
+
+fn parse_mwixnet_response(response: &str) -> Result<MwixnetResponse, Error> {
```
ardocrat: can we move mwixnet related parsing to mwixnet file, it can be useful to use them at owner rpc api for example, same related to tests below

libwallet/src/mwixnet/onion/mod.rs
```diff
@@ -34,6 +34,9 @@ use x25519_dalek::{SharedSecret, StaticSecret};
 use crypto::secp::random_secret;
 use onion::{new_stream_cipher, Onion, OnionError, Payload, RawBytes};
 
+/// Maximum number of servers in an mwixnet route.
+pub const MAX_MWIXNET_HOPS: usize = 16;
```
ardocrat: Why such limit is important?

controller/src/command.rs
wiesche89: mwixnet swaps an existing UTXO, so reusing send would create an unnecessary Slate transaction. I can instead make the commitment optional and select an eligible output by amount or max, while keeping explicit selection available.

wiesche89: Added: Select an output by commitment, minimum amount, or max

controller/src/command.rs
wiesche89: Done with commit https://github.com/mimblewimble/grin-wallet/pull/783/commits/af1e0d040760826b3e1436ae656f85683567dcca

libwallet/src/mwixnet/onion/mod.rs
wiesche89: This limit is aligned with the upcoming route discovery protocol. Up to 8 hops, including the swap server. Since every hop adds a fee, longer routes would become unnecessarily expensive. Once route discovery is integrated, the wallet will use mwixnet_protocol::MAX_ROUTE_HOPS directly.

Added 8 hops in commit https://github.com/mimblewimble/grin-wallet/pull/783/commits/0cbf7a68602022bcf4f71fbf5b4af75da00410f3

Issue #662, opened 2022-07-28
Create single point of change for grin library imports


yeastplume:
As described in #661, it would be very useful to use whatever cargo tools are available to make version/branch changes for grin library imports as needed. At the moment, if you're switching to a development environment or bumping versions for a build, changes need to be made in each and every crate. It would be really useful to be able to do this in one place and have it propagate through all crates.

PR #787, opened 2026-08-17
centralize grin workspace dependencies

wiesche89:
**summary**

centralizes all grin dependencies in the root Cargo.toml using [workspace.dependencies].
this removes duplicated dependency definitions across the wallet crates and makes future path, branch, or version updates possible in one place.

addresses Issue : #662

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,14 @@ path = "src/bin/grin-wallet.rs"
 members = ["api", "config", "controller", "impls", "libwallet", "util"]
 exclude = ["integration"]
 
+[workspace.dependencies]
+grin_api = { path = "grin/api" }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -19,6 +19,14 @@ path = "src/bin/grin-wallet.rs"
 members = ["api", "config", "controller", "impls", "libwallet", "util"]
 exclude = ["integration"]
 
+[workspace.dependencies]
```
ardocrat, line 22: Since we are using grin deps as paths now, would it be useful to add non-grin deps as well?

ardocrat: this can lead to api compatibility issues if we use such crates separately at different projects on integration, lets leave it as it is now

ardocrat (approved):

PR #754, opened 2026-06-03
Update contracts to current staging + fix #729

jwinterm:
Brings the contracts branch current with `staging` and fixes the contract-transaction failure in #729, plus a hardening pass. Original phyro/Yeastplume history preserved; `cargo test --workspace` green (52 binaries / 185 tests).

**Fixes #729** (`46999ce`), payment proofs made opt-in. The receiver's `contract sign` unconditionally built an invoice promise with no sender address available, failing with `NoSenderAddressProvided`. Verified end-to-end with a real mainnet contract round-trip (kernel `08274ee4463a100334c2c4094cb1d2c429769dafb4f26c7fd615e5bab26ffcb5ff `).

Also in this PR:
- Merge current `staging` in (grin 5.4.0 dep bump; only `Cargo.lock` conflicted).
- Fix `VersionedSlate::version()` (was reporting V5 slates as V4).
- Stop panicking on bad input: a malformed v5 payment proof, an inconsistent contract slate, or bad `contract` CLI args now return errors instead of crashing the wallet.
- Remove the dead/broken `SEC_KEY_FAKE` scrubbing. Nonce reuse is already prevented by deleting the signing context after signing, now verifies the delete happened.

**Slate version default.** `create_slatepack` currently emits **v5** for every slate (the contracts branch changed it from v4). v5 is required for contract proofs' memo/timestamp, but defaulting *all* sends to v5 means v4-only wallets can't read them. Right fix is "emit the lowest sufficient version" (v4 unless the slate needs v5 fields)?

Slatepack::mode now uses SlatepackMode. The serialized values remain unchanged.

PR #793, opened 2026-08-17
support kernel features when creating slates

### LINKED ISSUE #393: Add support for NRD kernels
Add support for the new [NRD kernel type](https://github.com/mimblewimble/grin-rfcs/pull/47) and signature structure in the wallet.

### Context
[April 14 dev meeting](https://github.com/mimblewimble/grin-pm/blob/master/notes/20200414-meeting-development.md):
> - _yeastplume:_ Actually, reminds me I had a question on nerd kernels just in respect to the wallet, should there be a separate issue to support them in the wallet, support for the new kernel type and new signature structure?
>    - _antiochp:_ Hmm - yeah actually _using_ them is non-trivial.
>    - _jaspervdm:_ I think regular send operations do not need to use them, so no for now.
>    - _antiochp:_ And not really much use outside multiparty outputs.
>    - _jaspervdm:_ Until we want to implement some form of payment channel.
>    - _yeastplume:_ Yes, I thought so, but I was thinking a bare minimum... Recognizing them and supporting the new sig format. With a more fully-fleshed out api to come much later. I suppose this can all come after the hf anyhow.
>    - _antiochp:_ No sig changes.
>    - _tromp:_ Wallet should just skip all kernels it doesn't recognize. E.g. when recovering.
>    - _antiochp:_ Oh I see what you mean, message being changed is specific to the kernel type.
>    - _yeastplume:_ Additional sig format? Or did I read that wrong. There should probably be some bare minimum recognition of them in the wallet, I don't think we'd want a state where some wallets in the wild know about them and others don't.
>    - _jaspervdm:_ Not sure I understand, any tx the wallet will construct will either have a plain kernel or a lock height kernel, it does not ever construct a nrd kernel.
>    - _antiochp:_ Hmm - want to open an issue on grin-wallet to track this conversation.
>    - _jaspervdm:_ Unless we explicitly want to add that feature.
>    - _antiochp:_ We can comment there offline.
>    - _jaspervdm:_ But not sure we want to.
>    - _yeastplume:_ But some other wallet might.
>    - _jaspervdm:_ Wallet could simply refuse to sign such a tx.
>    - _yeastplume:_ Anyhow, yes we can take this offline.
>    - _jaspervdm:_ Since it requires an additional fee. Yes agreed.

### LINKED ISSUE #564: Unable to set the `kernel_features_args` in Slate object

wiesche89:
Adds typed kernel feature support for new slates and fixes height locked V4 reconstruction.
Includes coverage for kernel creation and roundtrips.
Also preserves NRD kernel features through V4 Slatepack roundtrips

Fixes #564 
Fixes #393 
Replace: #565

### DIFF
--- a/libwallet/src/slate.rs
+++ b/libwallet/src/slate.rs
@@ -986,6 +986,8 @@ pub fn tx_from_slate_v4(slate: &SlateV4) -> Option<Transaction> {
 	let secp = secp.lock();
 	let mut calc_slate = Slate::blank(2, false);
 	calc_slate.fee_fields = slate.fee;
+	calc_slate.kernel_features = slate.feat;
+	calc_slate.kernel_features_args = slate.feat_args.as_ref().map(KernelFeaturesArgs::from);
 	for d in slate.sigs.iter() {
 		calc_slate.participant_data.push(ParticipantData {
 			public_blind_excess: d.xs,
@@ -1001,18 +1003,11 @@ pub fn tx_from_slate_v4(slate: &SlateV4) -> Option<Transaction> {
 		Ok(s) => s,
 		Err(_) => Signature::from_raw_data(&[0; 64]).unwrap(),
 	};
+	let features = calc_slate
+		.kernel_features()
+		.unwrap_or(KernelFeatures::Plain { fee: slate.fee });
 	let kernel = TxKernel {
-		features: match slate.feat {
-			0 => KernelFeatures::Plain { fee: slate.fee },
-			2 => KernelFeatures::HeightLocked {
-				fee: slate.fee,
-				lock_height: match slate.feat_args.as_ref() {
-					Some(a) => a.lock_hgt,
-					None => 0,
-				},
-			},
-			_ => KernelFeatures::Plain { fee: slate.fee },
-		},
+		features,
 		excess,
 		excess_sig,
 	};
--- a/libwallet/src/slate.rs
+++ b/libwallet/src/slate.rs
@@ -276,6 +276,38 @@ impl Slate {
 			kernel_features_args: None,
 		}
 	}
+
+	/// Create a new slate with the provided kernel features.
+	pub fn blank_with_kernel_features(
+		num_participants: u8,
+		is_invoice: bool,
+		kernel_features: KernelFeatures,
+	) -> Result<Slate, Error> {
+		let feature = kernel_features.as_u8();
+		let (fee_fields, kernel_features_args) = match kernel_features {
+			KernelFeatures::Plain { fee } => (fee, None),
+			KernelFeatures::Coinbase => return Err(Error::InvalidKernelFeatures(feature)),
+			KernelFeatures::HeightLocked { fee, lock_height } => {
+				(fee, Some(KernelFeaturesArgs { lock_height }))
+			}
+			KernelFeatures::NoRecentDuplicate {
+				fee,
+				relative_height,
+			} => (
+				fee,
+				Some(KernelFeaturesArgs {
+					lock_height: relative_height.into(),
+				}),
+			),
+		};
+		let mut slate = Slate::blank(num_participants, is_invoice);
+		slate.fee_fields = fee_fields;
+		slate.kernel_features = feature;
+		slate.kernel_features_args = kernel_features_args;
+		slate.update_kernel()?;
+		Ok(slate)
+	}
+
 	/// Removes any signature data that isn't mine, for compacting
 	/// slates for a return journey
 	pub fn remove_other_sigdata<K>(
@@ -972,7 +1004,7 @@ pub fn tx_from_slate_v4(slate: &SlateV4) -> Option<Transaction> {
 	let kernel = TxKernel {
 		features: match slate.feat {
 			0 => KernelFeatures::Plain { fee: slate.fee },
-			1 => KernelFeatures::HeightLocked {
+			2 => KernelFeatures::HeightLocked {
 				fee: slate.fee,
 				lock_height: match slate.feat_args.as_ref() {
 					Some(a) => a.lock_hgt,
--- a/libwallet/src/slate_versions/mod.rs
+++ b/libwallet/src/slate_versions/mod.rs
@@ -77,13 +77,22 @@ impl From<VersionedSlate> for Slate {
 	}
 }
 
-#[derive(Deserialize, Serialize)]
+#[derive(Serialize)]
 #[serde(untagged)]
 /// Binary versions, can only be parsed 1:1 into the appropriate
 /// version, and VersionedSlate can up/downgrade from there
 pub enum VersionedBinSlate {
 	/// Version 4, binary
 	V4(SlateV4Bin),
+}
+
+impl<'de> serde::Deserialize<'de> for VersionedBinSlate {
+	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+	where
+		D: serde::Deserializer<'de>,
+	{
+		<SlateV4Bin as serde::Deserialize>::deserialize(deserializer).map(Self::V4)
+	}
 }
 
 impl TryFrom<VersionedSlate> for VersionedBinSlate {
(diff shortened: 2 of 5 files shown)


libwallet/src/slate.rs
```diff
@@ -969,18 +1003,11 @@ pub fn tx_from_slate_v4(slate: &SlateV4) -> Option<Transaction> {
 		Ok(s) => s,
 		Err(_) => Signature::from_raw_data(&[0; 64]).unwrap(),
 	};
+	let features = calc_slate
+		.kernel_features()
+		.unwrap_or(KernelFeatures::Plain { fee: slate.fee });
```
wiesche89, line 1008: I'll change this to .ok()? so invalid or incomplete kernel features do not silently fall back to a plain kernel.

libwallet/src/slate_versions/v4_bin.rs
```diff
@@ -31,6 +31,10 @@ use crate::slate_versions::v4::{
 	VersionCompatInfoV4,
 };
 
+// KernelFeatures constants from grin_core
```
ardocrat, line 34: seems like its needed to make this constants public at grin_core to not duplicate

libwallet/src/slate_versions/v4_bin.rs
```diff
@@ -475,9 +479,22 @@ impl Readable for SlateV4Bin {
 		let sigs = SigsWrap::read(reader)?.0;
 		let opt_structs = SlateOptStructs::read(reader)?;
 
-		let feat_args = if opts.feat == 2 {
+		let feat_args = if matches!(opts.feat, HEIGHT_LOCKED_U8 | NO_RECENT_DUPLICATE_U8) {
 			Some(KernelFeaturesArgsV4 {
-				lock_hgt: reader.read_u64()?,
+				lock_hgt: reader.read_u64().map_err(|err| {
```
ardocrat, line 484: Do you think we should also validate NRDRelativeHeight value here to be over 0 and less than MAX (consensus::WEEK_HEIGHT)? It will require to make MAX field public here at grin_core as above.

libwallet/src/slate_versions/v4_bin.rs
ardocrat: Made constants public at https://github.com/mimblewimble/grin/pull/3935

libwallet/src/slate_versions/v4_bin.rs
ardocrat: Made constants public at https://github.com/mimblewimble/grin/pull/3935, will add check

libwallet/src/slate.rs
ardocrat: also added error logging

libwallet/src/slate.rs
```diff
@@ -954,33 +982,30 @@ pub fn tx_from_slate_v4(slate: &SlateV4) -> Option<Transaction> {
 	let secp = secp.lock();
 	let mut calc_slate = Slate::blank(2, false);
 	calc_slate.fee_fields = slate.fee;
+	calc_slate.kernel_features = slate.feat;
+	calc_slate.kernel_features_args = slate.feat_args.as_ref().map(KernelFeaturesArgs::from);
```
wiesche89, line 986: This can now reconstruct NRD transactions, but grin-wallet never initializes the global NRD flag. Loading a stored NRD transaction will therefore fail in core deserialization and panic at the unwrap() in get_stored_tx(). Please initialize the flag for the selected network, propagate the error instead of unwrapping it, and add a store/load test. The same fix is needed in [#754](https://github.com/mimblewimble/grin-wallet/pull/754)

libwallet/src/slate_versions/v4_bin.rs
```diff
@@ -443,16 +443,30 @@ impl Writeable for SlateV4Bin {
...
 			coms: &v4.coms,
 			proof: &v4.proof,
 		}
 		.write(writer)?;
-		// Write lock height for height locked kernels
-		if v4.feat == 2 {
+		// Write the height argument for height locked and NRD kernels
+		if matches!(
+			v4.feat,
+			KernelFeatures::HEIGHT_LOCKED_U8 | KernelFeatures::NO_RECENT_DUPLICATE_U8
```
wiesche89, line 455: This adds the NRD height to the existing V4 binary format without changing its version. Older V4 readers ignore the extra field and lose feat_args; some paths may then rebuild the kernel as Plain. [#754](https://github.com/mimblewimble/grin-wallet/pull/754) 
can still select V4 for NRD, so it does not fully solve this. Please include kernel features in the version/capability selection.

libwallet/src/slate_versions/v4_bin.rs
```diff
@@ -443,16 +443,30 @@ impl Writeable for SlateV4Bin {
...
+		) {
 			let lock_hgt = match &v4.feat_args {
-				Some(l) => l.lock_hgt,
+				Some(l) => {
+					// Check valid NRD Slatepack relative height.
+					if v4.feat == KernelFeatures::NO_RECENT_DUPLICATE_U8
```
wiesche89: There still seems to be a case where we can create a slatepack that our own reader rejects when the NRD height is missing.

Could we validate the height in both cases and add a test for an invalid height as well? Also, the more specific error gets replaced by SlatepackSer, so it would be nice to keep the actual reason.

libwallet/src/slate_versions/mod.rs
```diff
@@ -86,6 +86,15 @@ pub enum VersionedBinSlate {
 	V4(SlateV4Bin),
 }
 
+impl<'de> serde::Deserialize<'de> for VersionedBinSlate {
+	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+	where
+		D: serde::Deserializer<'de>,
+	{
+		<SlateV4Bin as serde::Deserialize>::deserialize(deserializer).map(Self::V4)
```
wiesche89, line 94: This always tries to deserialize the binary slate as V4. That conflicts with contract branch, which supports both V4 and V5. Please dispatch based on the declared version while preserving the specific deserialization error.

libwallet/src/slate_versions/v4_bin.rs
```diff
@@ -443,16 +443,30 @@ impl Writeable for SlateV4Bin {
...
+						&& NRDRelativeHeight::try_from(l.lock_hgt).is_err()
+					{
+						return Err(grin_ser::Error::IOErr(
+							"Invalid NRD Slatepack relative height".to_string(),
+							std::io::ErrorKind::InvalidData,
+						));
+					}
+					l.lock_hgt
+				}
 				None => 0,
```
wiesche89: Missing feat_args are written as height 0, even though 0 is invalid for NRD. This means the writer can create a slate that the reader rejects. Please return an error for missing required arguments and add writer/reader tests. Contracts branch has the same issue in V4 and V5.

ardocrat: done, seems like need to modify contracts branch too after merge this PR to staging

libwallet/src/slate_versions/mod.rs
ardocrat: it means we should add v5 and check type, can we add this after merge this PR to staging and later merge staging to contracts branch?

PR #795, opened 2026-08-17
fix tx filter

### LINKED ISSUE #789: Fix max_creation_timestamp filtering
max_creation_timestamp currently has no effect. The filter uses min_confirmed_timestamp when comparing creation_ts, which can also cause unexpected results for confirmation time queries.

Use max_creation_timestamp for the upper creation time bound and add a reg. test for both timestamp fields.

wiesche89:
Fixes max_creation_timestamp filtering and adds tests for both timestamp filters.

Fixes #789

### DIFF
--- a/controller/tests/tx_list_filter.rs
+++ b/controller/tests/tx_list_filter.rs
@@ -61,6 +61,25 @@ fn test_wallet_tx_filtering(
 		mask,
 		PathBuf::from(test_dir),
 		|api, _m| {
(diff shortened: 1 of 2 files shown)


ardocrat (approved):
Seems like it was bug in original implementation https://github.com/mimblewimble/grin-wallet/pull/666 as `max_creation_timestamp` filter missed at all.

PR #799, opened 2026-08-17
reject zero send amounts

### LINKED ISSUE #511: Incorrect error messagen when sending 0 grin (Nit-Pick)
**Describe the bug**
When sending 0 grin the error is "Not enough funds. Required: 0.005000000, Available: 0.000000000"

**To Reproduce**
From a wallet that has some spendable grin, send 0 grin to another wallet.

**Expected behavior**
Error message should say "Invalid send amount" or something.
or, Maybe it should allow sending 0 grin as long as the wallet has enough to cover the fee?

**Screenshots**
```
grin@crypto:~$ grin-wallet send -d grin1xxxyyyzzz 0
Password:
Wallet command failed: LibWallet Error: Not enough funds. Required: 0.005000000, Available: 0.000000000
```

wiesche89:
Return a clear InvalidAmount error when attempting to send zero GRIN instead of reporting insufficient funds.

Fixes #511

Also avoid of creation invoice with 0 amount.

### DIFF
--- a/libwallet/src/api_impl/owner.rs
+++ b/libwallet/src/api_impl/owner.rs
@@ -530,6 +530,10 @@ where
 	C: NodeClient,
 	K: Keychain,
 {
+	if args.amount == 0 {
+		return Err(Error::InvalidAmount);
+	}
+
 	let payment_proof_address = if let Some(a) = &args.payment_proof_recipient_address {
 		if a.valid_network() {
 			Some(a)
--- a/libwallet/src/error.rs
+++ b/libwallet/src/error.rs
@@ -24,6 +24,10 @@ use grin_store;
 /// Wallet errors, mostly wrappers around underlying crypto or I/O errors.
 #[derive(Clone, Eq, PartialEq, Debug, thiserror::Error, Serialize, Deserialize)]
 pub enum Error {
+	/// Invalid transaction amount
+	#[error("Invalid amount (must be greater than zero)")]
+	InvalidAmount,
+
 	/// Not enough funds
 	#[error("Not enough funds. Required: {needed_disp:?}, Available: {available_disp:?}")]
 	NotEnoughFunds {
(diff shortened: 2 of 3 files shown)


libwallet/src/api_impl/owner.rs
```diff
@@ -530,6 +530,10 @@ where
 	C: NodeClient,
 	K: Keychain,
 {
+	if args.amount == 0 {
```
ardocrat, line 533: I'll add check to invoice creation too

libwallet/src/error.rs
```diff
@@ -24,6 +24,10 @@ use grin_store;
 /// Wallet errors, mostly wrappers around underlying crypto or I/O errors.
 #[derive(Clone, Eq, PartialEq, Debug, thiserror::Error, Serialize, Deserialize)]
 pub enum Error {
+	/// Invalid transaction amount
+	#[error("Invalid amount (must be greater than zero)")]
+	InvalidAmount,
```
ardocrat, line 29: I would also add this check at `foreign::receive_tx` and `owner::process_invoice_tx`.

libwallet/src/api_impl/owner.rs
wiesche89, line 533: With amount_includes_fee, an amount equal to the fee still produces a zero amount slate that the receiver rejects. Please catch this on the sender side before saving the context and add a test.

libwallet/src/api_impl/owner.rs
```diff
@@ -530,6 +530,10 @@ where
...
+		return Err(Error::InvalidAmount);
```
wiesche89, line 534: With send max on an empty wallet, this reports “Invalid amount”. Could we say “No spendable funds” instead in the cli and add a test?

ardocrat: added

libwallet/src/api_impl/owner.rs
ardocrat: Added

ardocrat (approved):

PR #800, opened 2026-08-17
clarify bip32 derivation paths

### LINKED ISSUE #663: wallet shows incorrect BIP32 derivation paths
**Describe the bug**

The key derivation path shown after running `grin-wallet account` is not correct.

**To Reproduce**
1. Create bunch of accounts
2. Run `grin-wallet account` to list them, note the derivation paths
3. Derive the slatepack addresses from those exact paths using GRIN++ or mimblewimble-py
4. Observe they do not match

**Expected behavior**

When I run

```bash
$ grin-wallet account
Password: 

____ Wallet Accounts ____

 Name    | Parent BIP-32 Derivation Path 
---------+-------------------------------
 alice   | m/1/0 
 bob     | m/2/0 
 default | m/0/0 

Command 'account' completed successfully
```

correct paths should be

```bash
$ grin-wallet account
Password: 

____ Wallet Accounts ____

 Name    | Parent BIP-32 Derivation Path 
---------+-------------------------------
 alice   | m/1/1/0
 bob     | m/2/1/0
 default | m/0/1/0

Command 'account' completed successfully
```

You can get the test data [here](https://github.com/marekyggdrasil/mimblewimble-py/blob/5379aaf3039139bc184cc5ed2aad22b6ece1108c/tests/test_wallet.py#L15-L44)

**Screenshots**

```bash
$ grin-wallet -V
grin-wallet 5.0.3

```

**Additional context**

wiesche89:
Shows the output parent path and Slatepack address path separately in the account list.

Fixes #663

### DIFF
--- a/controller/src/display.rs
+++ b/controller/src/display.rs
@@ -16,7 +16,8 @@ use crate::core::core::FeeFields;
 use crate::core::core::{self, amount_to_hr_string};
 use crate::core::global;
 use crate::libwallet::{
-	AcctPathMapping, Error, OutputCommitMapping, OutputStatus, TxLogEntry, ViewWallet, WalletInfo,
+	address, AcctPathMapping, Error, OutputCommitMapping, OutputStatus, TxLogEntry, ViewWallet,
+	WalletInfo,
 };
 use crate::util::ToHex;
 use grin_wallet_util::OnionV3Address;
@@ -537,12 +538,15 @@ pub fn accounts(acct_mappings: Vec<AcctPathMapping>) {
 
 	table.set_titles(row![
 		mMG->"Name",
-		bMG->"Parent BIP-32 Derivation Path",
+		bMG->"Output BIP-32 Parent Path",
+		bMG->"Slatepack BIP-32 Path (Index 0)",
 	]);
 	for m in acct_mappings {
+		let slatepack_path = address::address_derivation_path(&m.path, 0).to_bip_32_string();
 		table.add_row(row![
 			bFC->m.label,
 			bGC->m.path.to_bip_32_string(),
+			bGC->slatepack_path,
 		]);
 	}
 	table.set_format(*prettytable::format::consts::FORMAT_NO_BORDER_LINE_SEPARATOR);
--- a/libwallet/src/address.rs
+++ b/libwallet/src/address.rs
@@ -21,6 +21,15 @@ use grin_keychain::{ChildNumber, Identifier, Keychain, SwitchCommitmentType};
 
 use crate::blake2::blake2b::blake2b;
 
+/// Derive a Slatepack address path from an output parent path and index
+pub fn address_derivation_path(parent_key_id: &Identifier, index: u32) -> Identifier {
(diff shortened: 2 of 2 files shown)


aglkm:
Could you show how the table looks after the fix?

wiesche89:
Before:
image

After:
image

Multiple accounts:
image

libwallet/src/address.rs
```diff
@@ -30,20 +39,24 @@ pub fn address_from_derivation_path<K>(
 where
 	K: Keychain,
 {
-	let mut key_path = parent_key_id.to_path();
-	// An output derivation for acct m/0
-	// is m/0/0/0, m/0/0/1 (for instance), m/1 is m/1/0/0, m/1/0/1
-	// Address generation path should be
-	// for m/0: m/0/1/0, m/0/1/1
-	// for m/1: m/1/1/0, m/1/1/1
```
aglkm, line 38: I find these comments useful, maybe we shouldn't delete them from the code?

aglkm:
Thank you, this looks good to me.
For your consideration, how about the following naming?
```
         |           BIP-32 Derivation Path
  Name   | -----------------------------------------------
         | Parent Output Key | Slatepack Address (Index 0)
---------+------------------------------------------------
 alice   | m/1/0             | m/1/1/0
 bob     | m/2/0             | m/2/1/0
 default | m/0/0             | m/0/1/0
```

libwallet/src/address.rs
wiesche89: Good point, I added the examples back.

wiesche89:
> Thank you, this looks good to me. For your consideration, how about the following naming?
> 
> ```
>          |           BIP-32 Derivation Path
>   Name   | -----------------------------------------------
>          | Parent Output Key | Slatepack Address (Index 0)
> ---------+------------------------------------------------
>  alice   | m/1/0             | m/1/1/0
>  bob     | m/2/0             | m/2/1/0
>  default | m/0/0             | m/0/1/0
> ```

I like the idea, but prettytable doesn’t handle grouped headers very well. I’d keep the current layout. 
I got it looking like this, but the code needed for it isn’t very clean:

image

or (the “Name” column could also be moved down one row.)

image

ardocrat:
> image

this looks good, seems like default account should be on top?

controller/src/display.rs
```diff
@@ -537,12 +538,15 @@ pub fn accounts(acct_mappings: Vec<AcctPathMapping>) {
 
 	table.set_titles(row![
 		mMG->"Name",
-		bMG->"Parent BIP-32 Derivation Path",
+		bMG->"Output BIP-32 Parent Path",
+		bMG->"Slatepack BIP-32 Path (Index 0)",
 	]);
 	for m in acct_mappings {
```
ardocrat, line 544: We can add active account on top of list at Owner API (also useful for GUI to avoid manual sort)

wiesche89: added

ardocrat (approved):

PR #801, opened 2026-08-17
avoid empty wallet seeds

### LINKED ISSUE #586: Lifecycle Error: Error recovering wallet seed
**Describe the bug**
I get this error when trying to get the recovery seed phrase for a wallet.

**To Reproduce**
This wallet was created using the wallets owner_api "create_wallet" call using the following parameters:
{"password": "supersecret", "name": null, "mnemonic": null, "mnemonic_length": 0}

The create_wallet call returns success and the wallet appears to work. I can scan it, and get "info", etc.

Getting the recovery phrase fails with:
```
grin@grinwallet:~$ grin-wallet recover
Password: 
Wallet command failed: LibWallet Error: Lifecycle Error: Error recovering wallet seed
```

**Expected behavior**
A recovery phrase should be returned and no error

**Additional context**
This is the seed in case it helps.  Password is **supersecret**
```
grin@grinwallet:~/.grin/main/wallet_data$ cat wallet.seed 
{
  "encrypted_seed": "f7dfdfbdf73faba595f43d530078a093",
  "salt": "693ff1f71d398171",
  "nonce": "deeaed12ae8443978e495977"
}
```

Note:  If I provide a valid "mnemonic" when calling create_wallet then the "grin-wallet recover" command works as expected.

wiesche89:
Uses a random 32 byte seed when mnemonic_length is 0 and returns a clearer recovery error for existing empty seed wallets.

Fixes #586

### DIFF
--- a/impls/src/lifecycle/default.rs
+++ b/impls/src/lifecycle/default.rs
@@ -196,6 +196,11 @@ where
 				return Err(Error::WalletSeedExists(msg));
 			}
 		}
+		let mnemonic_length = if mnemonic_length == 0 {
+			32
+		} else {
+			mnemonic_length
+		};
 		WalletSeed::init_file(
 			&data_dir_name,
 			mnemonic_length,
@@ -286,7 +291,7 @@ where
 			.map_err(|_| Error::Lifecycle("Error opening wallet seed file".into()))?;
 		let res = wallet_seed
 			.to_mnemonic()
-			.map_err(|_| Error::Lifecycle("Error recovering wallet seed".into()))?;
+			.map_err(|_| Error::Lifecycle("Wallet was created without a recovery phrase".into()))?;
 		Ok(ZeroingString::from(res))
 	}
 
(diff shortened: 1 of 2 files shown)


impls/src/lifecycle/default.rs
```diff
@@ -286,7 +291,7 @@ where
 			.map_err(|_| Error::Lifecycle("Error opening wallet seed file".into()))?;
 		let res = wallet_seed
 			.to_mnemonic()
-			.map_err(|_| Error::Lifecycle("Error recovering wallet seed".into()))?;
+			.map_err(|_| Error::Lifecycle("Wallet was created without a recovery phrase".into()))?;
```
ardocrat, line 294: Seems like we should return actual error and add more validation, `from_entropy` returns an error with invalid entropy size (mnemonic length), possible sizes are `[16, 20, 24, 28, 32]`.

ardocrat: Added returning of actual error (entropy length for `from_entropy`).

impls/src/lifecycle/default.rs
```diff
@@ -196,6 +196,11 @@ where
 				return Err(Error::WalletSeedExists(msg));
 			}
 		}
+		let mnemonic_length = if mnemonic_length == 0 {
```
wiesche89, line 199: The error is clearer now, but the length check is still missing. Please reject invalid BIP39 lengths before saving the seed, short seeds can be unsafe, and invalid lengths cannot produce a recovery phrase.

impls/src/lifecycle/default.rs
```diff
@@ -286,7 +291,7 @@ where
 			.map_err(|_| Error::Lifecycle("Error opening wallet seed file".into()))?;
 		let res = wallet_seed
 			.to_mnemonic()
-			.map_err(|_| Error::Lifecycle("Error recovering wallet seed".into()))?;
+			.map_err(|e| Error::Lifecycle(format!("Error recovering wallet seed: {}", e)))?;
```
wiesche89, line 294: An empty seed means the wallet is unsafe. Please warn users to move their funds to a new wallet, while keeping the old one accessible so they can do so.

ardocrat: A reason can be wrong entropy length, it means wallet is unusable, should we warn user to create new wallet and delete current?

impls/src/lifecycle/default.rs
ardocrat: added error clarification on create

impls/src/lifecycle/default.rs
ardocrat: added error clarification

ardocrat (approved):

PR #802, opened 2026-08-17
Add confirmed transaction height

### LINKED ISSUE #362: Refine `Confirmation Time` column in `grin-wallet txs`

wiesche89:
Stores and displays the block height where a transaction was confirmed, while keeping the local detection time.
Includes API, wallet and reorg coverage.

Fixes #362


libwallet/src/api_impl/owner.rs
```diff
@@ -1365,10 +1368,10 @@ where
 	};
 
 	for tx in txs.iter_mut() {
-		if tx.confirmed {
+		if tx.confirmed && tx.confirmed_height.is_some() {
 			continue;
 		}
-		if tx.amount_debited != 0 && tx.amount_credited != 0 {
+		if tx.amount_debited != 0 && tx.amount_credited != 0 && tx.confirmed_height.is_some() {
```
ardocrat, line 1374: Is it really possible to have `confirmed == false`, but `confirmed_height != None`?

wiesche89: You’re right, normal wallet updates don’t produce that state. I’ll adjust the condition

libwallet/src/types.rs
```diff
@@ -645,6 +645,18 @@ pub struct TxLogEntry {
...
 	pub confirmed: bool,
+	/// Observed confirmation height
+	/// Older entries may have no height
+	/// Known kernels are rechecked within the reorg window
+	#[serde(default)]
+	pub confirmed_height: Option<u64>,
+	/// Last known kernel height for bounded rechecks
+	#[serde(
+		default,
+		alias = "reverted_height",
```
ardocrat: Seems like not needed

wiesche89: removed

libwallet/src/api_impl/owner.rs
```diff
@@ -1338,6 +1346,77 @@ where
 	}
 }
 
+/// Recheck known heights, preserving status and detection time
+fn update_confirmed_heights<'a, L, C, K>(
```
ardocrat, line 1350: it can lead to a lot calls to node with a lot txs on every wallet status update, maybe we can detect if reorg really happened and ask for new kernels only once?

PR #804, opened 2026-08-18
use current header version in new slates

### LINKED ISSUE #525: incomplete HeaderVersion testing
libwallet/src/internal/tx.rs recognizes only HeaderVersion 1..3 and not 4?
with HF4 code, we should be able to just 
```
slate.version_info.block_header_version = header_version(current_height)
```
since every height will have a well-defined header version.

wiesche89:
Uses the header version for the current chain height when creating a slate instead of only handling versions 1 to 3.
Adds a test for the value.

Closes #525

### DIFF
--- a/libwallet/src/internal/tx.rs
+++ b/libwallet/src/internal/tx.rs
@@ -18,8 +18,7 @@ use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use std::io::Cursor;
 use uuid::Uuid;
 
-use crate::grin_core::consensus::valid_header_version;
-use crate::grin_core::core::HeaderVersion;
+use crate::grin_core::consensus::header_version;
 use crate::grin_keychain::{Identifier, Keychain};
 use crate::grin_util::secp::key::SecretKey;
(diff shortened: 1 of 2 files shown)


libwallet/src/internal/tx.rs
```diff
@@ -70,17 +69,7 @@ where
...
-	}
-
-	if valid_header_version(current_height, HeaderVersion(2)) {
-		slate.version_info.block_header_version = 2;
-	}
-
-	if valid_header_version(current_height, HeaderVersion(3)) {
-		slate.version_info.block_header_version = 3;
-	}
+	slate.version_info.block_header_version = header_version(current_height).0;
```
ardocrat, line 72: We also need to modify or get rid of `GRIN_BLOCK_HEADER_VERSION`, it has value of `3` and we set it on creation of blank state. Also there is weird hardcode of versions at `controller::check_middleware`:

Image

wiesche89: flip check

ardocrat (approved):

PR #805, opened 2026-08-18
unpack plain slatepacks without password

### LINKED ISSUE #469: Is wallet password needed for "unpack"?
If running "grin-wallet unpack" on a unencrypted slatepack does not need to use the wallet seed, please consider not requiring the wallet password for that function.

```
grinfloo@crypto:~$ grin-wallet --floonet unpack
Password: 
```

wiesche89:
Allows unencrypted Slatepacks to be unpacked without a wallet password. Encrypted Slatepacks still require the wallet to be opened.

Closes #469

### DIFF
--- a/src/cmd/wallet_args.rs
+++ b/src/cmd/wallet_args.rs
@@ -1058,12 +1058,26 @@ where
 	// local wallet proxy, etc.)
 	wallet_inst_cb(wallet.clone());
 
+	let unpack_slatepack = match wallet_args.subcommand() {
+		("unpack", Some(args)) => {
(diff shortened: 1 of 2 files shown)


src/cmd/wallet_args.rs
```diff
@@ -1058,12 +1058,26 @@ where
 	// local wallet proxy, etc.)
 	wallet_inst_cb(wallet.clone());
 
+	let unpack_slatepack = match wallet_args.subcommand() {
```
ardocrat: we can handle this command below to not duplicate and avoid None

PR #807, opened 2026-08-20
reject wrong slate states in receive and finalize

### LINKED ISSUE #439: Improve error message when invoice is processed with incorrect command
Currently, the wallet has a `pay` and `receive` command. It's somewhat unclear by command name alone which one is the expected command to be used when the user is responding to an invoice.

The correct command is `pay`. Incorrectly using `receive` does not return a descriptive error message. This could be improved by pointing the user to `pay` in order to complete the workflow.

#### Invoice issuer: v4.0.0
1. grin-wallet v4.0.0-beta2 on linux.
2. Command:

   ```
   $ ./grin-wallet invoice 0.392
   Password: 

   Transaction file was output to:

   /home/lehnberg/.grin/main/slatepack/b37ac032-8ec2-4e67-8f67-61b3dcb0a227.I1.slatepack
   ```

#### Invoice recipient: v3.0.0
1. grin-wallet v3.0.0 on linux.
2. Command:
   ```
   $ ./grin-wallet receive -i ~/b37ac032-8ec2-4e67-8f67-61b3dcb0a227.I1.slatepack
   Password: 
   Wallet command failed: LibWallet Error: Fee Error: Fee Dispute Error: 0, 9000000
   ```

## Expected result
v3.0.0 wallet should detect that this is an invoice flow slate, and therefore reject the use of `receive` command, and instead output an error message directing the user to process the invoice using `pay`.

## Actual result
v3.0.0 wallet fails with `LibWallet Error: Fee Error: Fee Dispute Error: 0, 9000000`, and treats the transaction as if it was money it was due to receive (see comment below).

wiesche89:
Rejects slates passed to "receive_tx" that are not in state S1. The "receive" and "finalize" commands tell the user which command to use instead, e.g. "pay" for invoice slates.

This prevents receive from converting an invoice slate to S2 and creating an incorrect received transaction and output.

Fixes #439

### DIFF
--- a/libwallet/src/api_impl/foreign.rs
+++ b/libwallet/src/api_impl/foreign.rs
@@ -66,6 +66,9 @@ where
 	K: Keychain,
 {
 	let mut ret_slate = slate.clone();
+	if ret_slate.state == SlateState::Invoice1 {
+		return Err(Error::InvoiceSlateRequiresPay);
+	}
 	check_ttl(w, &ret_slate)?;
 	let parent_key_id = match dest_acct_name {
 		Some(d) => {
--- a/libwallet/src/error.rs
+++ b/libwallet/src/error.rs
@@ -217,6 +217,10 @@ pub enum Error {
 	#[error("Invalid slate state")]
 	SlateState,
 
+	/// Invoice passed to receive
+	#[error("Use the 'pay' command to process invoice slates")]
+	InvoiceSlateRequiresPay,
+
 	/// Can't serialize slate pack
 	#[error("Can't Serialize slatepack")]
 	SlatepackSer,
--- a/controller/tests/invoice.rs
+++ b/controller/tests/invoice.rs
@@ -124,6 +124,27 @@ fn invoice_tx_impl(test_dir: &'static str) -> Result<(), libwallet::Error> {
 	)?;
 	assert_eq!(slate.state, SlateState::Invoice1);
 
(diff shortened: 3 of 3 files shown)


libwallet/src/api_impl/foreign.rs
```diff
@@ -66,6 +66,9 @@ where
 	K: Keychain,
 {
 	let mut ret_slate = slate.clone();
+	if ret_slate.state == SlateState::Invoice1 {
```
ardocrat: Seems like we need to handle invoice2 too, if tx not exists it shows `Wallet command failed: LibWallet Error: LibTx Error` for existing tx it shows `Wallet command failed: LibWallet Error: Transaction has already been received` (whats wrong, cause it has invoice1 pending status). At this case it requires `finalize` command. I would return `Error::SlateState` for all states except `Standard1` to show right message only at CLI just after parse.

libwallet/src/error.rs
```diff
@@ -217,6 +217,10 @@ pub enum Error {
 	#[error("Invalid slate state")]
 	SlateState,
 
+	/// Invoice passed to receive
+	#[error("Use the 'pay' command to process invoice slates")]
```
ardocrat: We also use this method also at foreign API when sharing address, so such message should be handled at different place only for CLI calls, also for all non-`Standard1` states.

ardocrat (approved):

Issue #813, opened 2026-09-15
Lock outputs before Tor send


wiesche89:
init_send_tx locks outputs only after a successful Tor send. During the request, parallel sends can select the same outputs.
We should lock them before calling Tor and keep them locked on failure, since the slate can still be sent manually. cancel_tx can unlock them if needed.

PR #814, opened 2026-09-16
Lock outputs before sending with Tor

ardocrat:
Lock outputs before sending with Tor on CLI `send` command and owner API call with `send_args`.

Closes https://github.com/mimblewimble/grin-wallet/issues/813


api/src/owner.rs
```diff
@@ -806,11 +806,11 @@ where
 				if self.doctest_mode || !can_send || dest.is_none() {
 					return Ok(slate);
 				}
+				self.tx_lock_outputs(keychain_mask, &slate)?;
```
wiesche89, line 809: 1) Selection and locking still happen in two separate wallet-lock sections. A concurrent request can select the same outputs in between, and lock_tx_context() does not revalidate their current status. In an interleaved test, both transactions were therefore able to lock the same outputs successfully. Please make selection and locking atomic for auto-send, or reject stale selections when locking, and add a test for this sequence.

2) With send_args, init_send_tx() now locks the outputs itself, even when the Tor send fails. However, the documentation still tells callers to invoke tx_lock_outputs() afterwards, which can create a second TxSent entry. Please update the documentation and example for the new behavior,

controller/src/command.rs
```diff
@@ -491,30 +490,31 @@ where
...
 		)?)
 	};
 
 	let can_send = tor_config.send_tor(args.skip_tor);
 	if test_mode || !can_send || dest.is_none() {
-		return output_sp();
+		return output_sp(owner_api, true);
 	}
 
+	owner_api.tx_lock_outputs(keychain_mask, &slate)?;
```
wiesche89, line 511: With late_lock, the context does not contain any inputs yet. they are only selected during finalization. This call therefore locks nothing but still creates an empty TxSent entry. If Tor fails, that misleading entry is left behind. Please handle late_lock explicitly here and test the failure path. The same applies to the new call in the Owner API.

controller/src/command.rs
```diff
@@ -529,7 +529,7 @@ where
 		}
 		Err(e) => {
 			error!("Error sending slate sync: {}", e);
-			output_sp()?;
+			output_sp(owner_api, false)?;
```
wiesche89, line 532: After a failed Tor send, the outputs now intentionally remain locked. The error output only prints the slate, however, and does not explain that it must be sent manually or that the transaction must be cancelled to unlock the outputs. Please include that guidance in this failure path.

ardocrat: It writes as `Slatepack data follows. Please provide this output to the other party` should we also provide instruction (command) how to cancel?

api/src/owner.rs
ardocrat: fixed by using same wallet lock for selecting and locking

controller/src/command.rs
ardocrat: fixed

controller/src/command.rs
ardocrat: added mention that tx can be canceled

controller/src/command.rs
ardocrat: added tests

api/src/owner.rs
```diff
@@ -785,16 +786,15 @@ where
...
-			let tor_config = send_args
-				.as_ref()
-				.map(|_| crate::tor_config::load(&self.config_path()))
-				.transpose()?;
-			let w = w_lock.lc_provider()?.wallet_inst()?;
-			let slate = owner::init_send_tx(w, keychain_mask, args, self.doctest_mode)?;
-			(slate, tor_config)
-		};
+
+		wallet_lock!(self.wallet_inst, w);
```
wiesche89: The wallet lock is still held when self.finalize_tx() tries to acquire it again, causing a deadlock. Please keep selection and output locking in one scope, then release the lock before the Tor call. A successful auto send test with and without late locking would catch this.

controller/tests/late_lock.rs
```diff
@@ -151,6 +152,29 @@ fn late_lock_test_impl(test_dir: &'static str) -> Result<(), libwallet::Error> {
...
+				late_lock: Some(true),
+				send_args: Some(InitTxSendArgs {
+					dest: "tgrin1xtxavwfgs48ckf3gk8wwgcndmn0nt4tvkl8a7ltyejjcy2mc6nfs9gm2lp".into(),
+					post_tx: false,
+					fluff: false,
+					skip_tor: Some(false),
+				}),
+				..Default::default()
+			};
+			let err = api.init_send_tx(m, args).err();
```
wiesche89: This also applies to tests/cmd_line_basic.rs:756.
The api test stops at a config error and rechecks the old balance, while the cli test skips Tor entirely. Please use a valid config, trigger a transport failure, and then reload the balance and transaction log an empty TxSent won’t change the balance.
For the parallel test, perhaps setting the chain type in each worker would be enough while sharing the wallet config between threads.

api/src/owner.rs
wiesche89: Point 2: The example is updated, but the main docs still say to call tx_lock_outputs() afterwards. After a failed auto send, this creates a duplicate TxSent and breaks cancellation by slate uuid. Please update the docs, including the skipped Tor and late lock cases.

controller/src/command.rs
wiesche89: With late locking, there’s no TxSent to cancel yet. Please skip that hint there. For normal send failures, explain that outputs stay locked and show cancel -t <uuid> with the actual uuid.

api/src/owner.rs
ardocrat: fixed

api/src/owner.rs
ardocrat: updated

controller/tests/late_lock.rs
ardocrat: fixed tests, added parallel test

controller/tests/transaction.rs
```diff
@@ -81,459 +85,419 @@ fn basic_transaction_api(test_dir: &'static str) -> Result<(), libwallet::Error>
...
+			minimum_confirmations: 2,
+			max_outputs: 500,
+			num_change_outputs: 1,
+			selection_strategy_is_use_all: true,
+			amount: 1,
+			send_args: Some(InitTxSendArgs {
+				dest: "tgrin1xtxavwfgs48ckf3gk8wwgcndmn0nt4tvkl8a7ltyejjcy2mc6nfs9gm2lp".into(),
+				post_tx: false,
+				fluff: false,
+				skip_tor: Some(false),
```
wiesche89: Could we trigger a controlled sender failure here, as in the late_lock test? This test still starts Arti and uses the Tor network. Skipping Tor would not work because the API would return before locking the outputs.

controller/src/controller.rs
```diff
@@ -136,23 +136,6 @@ fn init_tor_listener(
 	Ok(process)
 }
 
-/// Instantiate wallet Owner API for a single-use (command line) call
-/// Return a function containing a loaded API context to call
-pub fn owner_single_use<L, F, C, K>(
```
wiesche89: Could we keep owner_single_use as a compatibility wrapper? Removing this public function breaks library callers that use it, and the locking fix does not require its removal.

api/src/owner.rs
wiesche89: The docs still include manual forwarding after a failed Tor send, when outputs are already locked. Calling tx_lock_outputs again creates a second TxSent and breaks cancellation by UUID. Please clarify when callers need to lock, including skipped sending and late locking.

controller/tests/late_lock.rs
wiesche89: The sender failure is now controlled, but wallet_info still comes from before the call. Please reload the balance and check that no TxSent exists for the returned slate. The CLI test still skips the failure path through test_mode, so it also needs a controlled failure and a transaction log check.

controller/src/command.rs
wiesche89: The hint still appears with late locking, when there is nothing to cancel yet. Please hide it there and show the cancellation command with the actual UUID for normal send failures. Checking the lock argument alone would miss those failures because their outputs are already locked.

Issue #812, opened 2026-09-15
Add balances and Slatepack address index to account list


wiesche89:
- Show the balance of each account in the account list.
- Add a CLI option for the Slatepack address index. It's always 0 now, the Owner API already supports other indexes.

PR #815, opened 2026-09-17
Accounts balances

ardocrat:
Display balances at CLI and from owner API when minimum confirmations are set.

Partially solves https://github.com/mimblewimble/grin-wallet/issues/812


libwallet/src/types.rs
```diff
@@ -772,6 +772,9 @@ pub struct AcctPathMapping {
 	pub label: String,
 	/// Corresponding parent BIP32 derivation path
 	pub path: Identifier,
+	/// Wallet information for this account, can be loaded after deserialization.
+	#[serde(skip)]
```
wiesche89, line 776: serde(skip) also removes info from json rpc responses, so clients never receive the balances. Could we keep the stored account mapping separate from the api response and serialize the balance there?

src/bin/grin-wallet.yml
```diff
@@ -48,6 +48,12 @@ subcommands:
             short: c
             long: create
             takes_value: true
+        - minimum_confirmations:
+            help: Minimum number of confirmations required for an output to be spendable
+            short: c
```
wiesche89, line 53: --create already uses -c. This makes CLI construction panic with debug assertions enabled

libwallet/src/api_impl/owner.rs
```diff
@@ -52,12 +52,26 @@ use std::sync::mpsc::Sender;
...
 	C: NodeClient,
 	K: Keychain,
 {
-	let mut accounts = keys::accounts(w)?;
+	let mut keys_accounts = keys::accounts(w)?;
+	let mut accounts: Vec<AcctPathMapping> = if let Some(mc) = minimum_confirmations {
+		keys_accounts
+			.iter_mut()
+			.map(|a| {
+				a.info = updater::retrieve_info(w, &a.path, mc).ok();
```
wiesche89, line 68: Could we propagate this error instead of using .ok()? 
a database read failure currently looks the same as “balance not requested”, and the cli only shows -. The existing summary info path propagate these errors

libwallet/src/api_impl/owner.rs
```diff
@@ -52,12 +52,26 @@ use std::sync::mpsc::Sender;
...
 where
 	C: NodeClient,
 	K: Keychain,
 {
-	let mut accounts = keys::accounts(w)?;
+	let mut keys_accounts = keys::accounts(w)?;
+	let mut accounts: Vec<AcctPathMapping> = if let Some(mc) = minimum_confirmations {
+		keys_accounts
+			.iter_mut()
+			.map(|a| {
```
wiesche89, line 67: retrieve_info reads all wallet outputs, so this repeat the full database read for every account while holding the wallet lock. Could we share one read across accounts while keeping the existing balance rules and each account stored height?

libwallet/src/api_impl/owner.rs
```diff
@@ -52,12 +52,26 @@ use std::sync::mpsc::Sender;
...
+				a.info = updater::retrieve_info(w, &a.path, mc).ok();
+				a.clone()
```
wiesche89, line 69: Could we update the existing account vector in place? We already own it, so cloning every entry into a second vector adds allocations without preserving anything we still need.

api/src/owner.rs
```diff
@@ -294,12 +295,12 @@ where
 	pub fn accounts(
 		&self,
 		keychain_mask: Option<&SecretKey>,
+		minimum_confirmations: Option<u64>,
```
wiesche89, line 298: Please document that None skips balance calculation and Some(n) uses the locally stored wallet state without refreshing from the node.

libwallet/src/types.rs
ardocrat: removed skip here, we already save accounts without info, added additional protection at save_acct_path

src/bin/grin-wallet.yml
ardocrat: changed to mc

libwallet/src/api_impl/owner.rs
ardocrat: added

libwallet/src/api_impl/owner.rs
ardocrat: fixed

api/src/owner.rs
ardocrat: added

libwallet/src/api_impl/owner.rs
ardocrat: added separate method to retrieve info for accounts

ardocrat (commented):
Added active account persistence over `Owner::set_active_account` and with CLI command `account --active "label"`

libwallet/src/backend.rs
```diff
@@ -275,27 +293,59 @@ where
...
+					let mut batch = self.batch(keychain_mask)?;
+					let current = batch
+						.acct_path_iter()?
+						.filter(|a| a.current.unwrap_or(false))
+						.collect::<Vec<AcctPathMapping>>();
+					for mut a in current {
+						a.current = None;
+						batch.save_acct_path(a.clone())?;
+					}
+					batch.commit()?;
```
wiesche89: Please clear the old flag and set the new one in a single batch. An interruption between commits makes the wallet reopen on default. Please cover this with a restart test.

libwallet/src/types.rs
```diff
@@ -768,10 +768,14 @@ impl ser::Readable for StoredProofInfo {
...
-	/// label used by user
+	/// Label used by user.
 	pub label: String,
-	/// Corresponding parent BIP32 derivation path
+	/// Corresponding parent BIP32 derivation path.
 	pub path: Identifier,
+	/// Wallet information for this account, can be loaded after deserialization.
+	pub info: Option<WalletInfo>,
+	/// Flag to check if this is current active account.
+	pub current: Option<bool>,
```
wiesche89: After a temporary account switch, current still marks the saved account. Older wallets mark none. Please derive this response field from the active path and test both cases.

controller/src/command.rs
```diff
@@ -1451,14 +1469,34 @@ where
...
+fn account_label<L, C, K>(
+	owner_api: &mut Owner<L, C, K>,
+	g_args: &GlobalArgs,
+) -> Result<String, Error>
+where
+	L: WalletLCProvider<'static, C, K> + 'static,
+	C: NodeClient + 'static,
+	K: keychain::Keychain + 'static,
+{
+	let label = if let Some(a) = g_args.account.as_ref() {
```
wiesche89: Starting the CLI with default, then switching to savings, still labels the savings address as default. receive also keeps the startup account. Please use the active account consistently after switching.

grin
```diff
@@ -1 +1 @@
-Subproject commit d5f32350ab548736c17c1db7ff51457917fe4e96
+Subproject commit 1425f603295ba166eccd9e4e0af0226742f90844
```
wiesche89: Please keep the existing Grin pin. The account changes build with it, so the extra PIHD and P2P changes are not needed here.

api/src/owner.rs
```diff
@@ -399,7 +411,7 @@ where
 		let w = w_lock.lc_provider()?.wallet_inst()?;
 		// Test keychain mask, to keep API consistent
 		let _ = w.keychain(keychain_mask)?;
-		owner::set_active_account(w, label)
+		owner::set_active_account(w, keychain_mask, label)
```
wiesche89: This now changes the selected account across restarts. Please mention that in the Rust and RPC docs, since existing clients could previously use this call for a temporary switch.

src/bin/grin-wallet.yml
```diff
@@ -48,6 +47,17 @@ subcommands:
             short: c
             long: create
             takes_value: true
+        - minimum_confirmations:
+            help: Minimum number of confirmations required for an output to be spendable
+            short: m
+            long: min_conf
+            default_value: "10"
+            takes_value: true
+        - active:
```
wiesche89: Combining create and active silently ignores the account selection. Please reject that combination and explain in the help which option saves the selection and which only overrides it temporarily.

src/cmd/wallet_args.rs
```diff
@@ -308,7 +308,11 @@ pub fn parse_global_args(
 	config: &WalletConfig,
 	args: &ArgMatches,
 ) -> Result<command::GlobalArgs, ParseError> {
-	let account = parse_required(args, "account")?;
+	let account = if args.subcommand_matches("account").is_some() {
```
wiesche89: Could we just use parse_optional(args, "account") here? The account command does not read this field, and open_wallet reads the original arguments, so this special case has no effect.

controller/src/display.rs
```diff
@@ -27,7 +27,7 @@ use term;
 
 /// Display outputs in a pretty way
 pub fn outputs(
-	account: &str,
+	account: &String,
```
wiesche89: Please keep &str here and in txs and info. The new String already works with it, so requiring &String only restricts callers.

